PR #34, opened 2025-04-24
Save Workflows

gitjannes:
Add the functionality to save a run as a standalone workflow to be used for future runs. 
This was called exporting a workflow in PROTzilla2, it is now renamed to saving.


pchklm:
These changes are on a new PR now: https://github.com/cschlaffner/PROTzilla/pull/51

PR #50, opened 2025-05-15
Improve test coverage

pchklm:
This PR makes sure only files that need testing are included into our coverage rate as well as adding new tests for multiple step calculation.


maximilianKalff (approved):
Quick Review, aber passt :)

PR #84, opened 2025-06-11
fixed husky

krohnjonas:
Husky now runs with our .git in the main folder. It will lint and format staged files before each commit.

To use it, you need to run pnpm prepare once. If you run pnpm install, this is done automatically, so it is included in our run-protzilla scripts.

To test whether the commit hooks are being executed, add 'exit 1' to the /frontend/.husky/pre-commit file and then try committing your changes. Your commit should then fail.
To test whether the linter runs, add an unused variable to an .ts file. The commit should then fail too.

### DIFF
--- a/frontend/.husky/pre-commit
+++ b/frontend/.husky/pre-commit
@@ -1,1 +1,2 @@
+cd frontend
 pnpx lint-staged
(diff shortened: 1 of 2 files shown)


maximilianKalff (approved):
Ultra cool :))

PR #105, opened 2025-11-14
Dependency update

hendraet:
## Description

Updates Python dependencies because some of them were quite outdated (and it was causing issues).  

## Changes

Changes code and test to account for changes in libraries. 

- numpy doesn't use `.NaN` anymore
- Isolation Forests can now work with nans
- GSEApy fixed Odds Ratio calc. in their API amongst others
- plus some minor stuff (mainly renamed parameters)


jorisfu:
@hendraet this should already be resolved with #106 , closing

PR #110, opened 2025-11-17
103 fix

### LINKED ISSUE #103: Fix "Skipping writing dataframe to x, valid file already exists" Warnings
## Issue description
Warnings like this occur in response to various actions in the software. However, they do not seem to have any impact on the UX

## Acceptance Criteria
- [ ] No more Warnings like this
- [ ] Nothing else changes once the warnings are gone 

## Still TODO
- [x] Use Mutex for dumping to prevent bad things from happening

jorisfu:
## Description
fixes #103

## Changes
For every step, PROTzilla now keeps track of a version number for each artifact (output or plot) generated by the step. The corresponding files on disk get overridden if and only if a new version was generated by a calculation previously. This improves performance for auto-save and removes the warnings described in #103 . It also fixes a bug where parameter re-adjustments within a step that was previously generated as valid would not lead to a disk rewrite of the step output.

A Mutex lock is also used for each disk saving operation to prevent race conditions from happening in the future.

## Testing
Create a new run, generate some artifacts and check via `ls -Rlah --full-time` whether or not all changes in PROTzilla that should lead to file changes also do so.

## PR checklist
**Development**
 
**Mergeability**
- [x] main-branch has been merged into local branch to resolve conflicts.
- [ ] The tests and linter have passed AFTER local merge (always fail right now, I propose we ignore that)
- [x] The frontend code has been formatted with `pnpm format` and checked with `pnpm lint` (no changes)
 
**Code review**

### DIFF
--- a/backend/protzilla/disk_operator.py
+++ b/backend/protzilla/disk_operator.py
@@ -71,11 +71,6 @@ class DataFrameOperator:
     @staticmethod
     def write(file_path: Path, dataframe: pd.DataFrame):
         with ErrorHandler():
-            if file_path.exists():
-                logger.warning(
-                    f"Skipping writing dataframe to {file_path}, valid file already exists"
-                )
-                return
             logger.info(f"Writing dataframe to {file_path}")
             dataframe.to_csv(file_path, index=False)
 
@@ -281,9 +276,6 @@ class DiskOperator:
                 step_data[KEYS.STEP_OUTPUTS] = self._write_output(step)
                 step_data[KEYS.STEP_MESSAGES] = step.messages.messages
                 step_data[KEYS.STEP_CALCULATION_STATUS] = step.calculation_status
-
-                # If step status is not "complete", reset dump state (definitely need to dump again)
-                step.is_dumped = (step.calculation_status == "complete")
             return step_data
 
     def _read_outputs(self, output: dict) -> Output:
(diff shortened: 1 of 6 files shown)


jorisfu:
Please ignore #5089c6c (formatting) and use #1b66853 as a diff reference for reviewing

backend/protzilla/disk_operator.py
```diff
@@ -311,19 +325,25 @@ def _read_plots(self, plots: dict) -> Plots:
...
                 self.plot_dir.mkdir(parents=True, exist_ok=True)
                 if not isinstance(
                     plot, bytes
                 ):  # TODO the data integration plots are of type byte, and therefore cannot be written using this methodology
                     write_json(plot, file_path)
                     plot.write_image(str(file_path).replace(".json", ".png"))
                     plots_data[i] = str(file_path)
-            return plots_data
+
+            self._update_dump_state(step, "plots")
```
Elena-kal, line 344: if _write_plots is supposed to return a dict, shouldn't we return plots_data? Right now we are not returning anything, right?

Elena-kal (changes requested):
Please check the comment I left for _write_plots(). The rest looks very reasonable.

jorisfu:
> Please check the comment I left for _write_plots(). The rest looks very reasonable.

Correct, should be fixed now

Elena-kal (approved):

Elena-kal:
Nice, the rest looks good :)

jorisfu:
#113 is fixed now, was caused by a problem in my code. @Elena-kal @3dot141592 can you check again if the bug is now gone for you as well?

AnnaPolensky (commented):
Looks good to me

AnnaPolensky (approved):

PR #112, opened 2025-11-18
Add Form definition to classes without form

### LINKED ISSUE #98: Fix "No Form defined" by migrating from Protzilla 2
## Issue description
For some steps (does anyone know which ones?) the error "no form defined" will show up. Hendrik fixed this for the Cluster, but for the other steps, we need to migrate the necessary code from Protzilla 2.

## Acceptance Criteria
- [ ] You can define a form, so that the "No Form defined"-Message won't show up.

## None-Goals
- [ ] Exeptions in the step are not handeled.
- [ ] Bugs and to-do's that where there before or in Protzilla2 are not fixed.

Elena-kal:
## Description
fixes #98
Adds forms to all classes without a form defined.

## Changes
All changes are made in [data_analysis.py](https://github.com/cschlaffner/PROTzilla/blob/form-analysis/backend/protzilla/methods/data_analysis.py). For every class that had its input defined create_form() and if needed a modify_form() were added. 
Other TODO's regarding the input flow are not solved with this pull request and are subject for another issue.

## Testing
Open Protzilla 2 (on the main branch) and Protzilla 1 on the form-definition branch. Then for each step in Data Analysis test wether the forms are working on Protzilla 1 (loading correctly and not throwing error in the terminal) and do not differ from the forms on Protzilla 2.

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**

### DIFF
--- a/backend/protzilla/methods/data_analysis.py
+++ b/backend/protzilla/methods/data_analysis.py
@@ -225,7 +225,7 @@ class DifferentialExpressionANOVA(DataAnalysisStep):
                     separatePrefix="\u03B1",
                 ),
                 DropdownField(name="grouping", label="Grouping from metadata"),
-                DropdownField(
+                MultiSelectField(
                     name="selected_groups", label="Select groups to perform ANOVA on"
                 ),
             ],
@@ -249,7 +249,9 @@ class DifferentialExpressionANOVA(DataAnalysisStep):
 
     def insert_dataframes(self, steps: StepManager, inputs) -> dict:
         inputs["log_base"] = steps.get_step_input(TransformationLog, "log_base")
-        inputs["intensity_df"] = steps.protein_df
+        inputs["intensity_df"] = steps.get_step_input(
+            Step, "protein_df", inputs["protein_df"]
+        )
         inputs["metadata_df"] = steps.metadata_df
         return inputs
 
@@ -355,7 +357,9 @@ class DifferentialExpressionTTest(DataAnalysisStep):
 
     def insert_dataframes(self, steps: StepManager, inputs) -> dict:
         inputs["log_base"] = steps.get_step_input(TransformationLog, "log_base")
-        inputs["intensity_df"] = steps.protein_df
+        inputs["intensity_df"] = steps.get_step_input(
+            Step, "protein_df", inputs["protein_df"]
+        )
         inputs["metadata_df"] = steps.metadata_df
         return inputs
 
@@ -714,7 +718,7 @@ class DifferentialExpressionKruskalWallisOnIntensity(DataAnalysisStep):
                     separatePrefix="\u03B1",
                 ),
                 DropdownField(name="grouping", label="Grouping from metadata"),
-                DropdownField(
+                MultiSelectField(
                     name="selected_groups",
                     label="Select groups to perform Kruskal-Wallis Test on",
                 ),
@@ -738,8 +742,11 @@ class DifferentialExpressionKruskalWallisOnIntensity(DataAnalysisStep):
     calc_method = staticmethod(kruskal_wallis_test_on_intensity_data)
 
     def insert_dataframes(self, steps: StepManager, inputs) -> dict:
-        inputs["ptm_df"] = steps.get_step_output(Step, "ptm_df", inputs["ptm_df"])
+        inputs["protein_df"] = steps.get_step_output(
+            Step, "protein_df", inputs["protein_df"]
+        )
         inputs["metadata_df"] = steps.metadata_df
+        inputs["log_base"] = steps.get_step_input(TransformationLog, "log_base")
         return inputs
 
 
@@ -779,7 +786,7 @@ class DifferentialExpressionKruskalWallisOnPTM(DataAnalysisStep):
                     separatePrefix="\u03B1",
                 ),
                 DropdownField(name="grouping", label="Grouping from metadata"),
-                DropdownField(
+                MultiSelectField(
                     name="selected_groups",
                     label="Select groups to perform Kruskal-Wallis Test on",
                 ),
@@ -1996,6 +2003,7 @@ class FLEXIQuantLF(DataAnalysisStep):
         )
 
         inputs["metadata_df"] = steps.metadata_df
+        return inputs
 
 
 class SelectPeptidesForProtein(DataAnalysisStep):
--- a/backend/protzilla/methods/data_analysis.py
+++ b/backend/protzilla/methods/data_analysis.py
@@ -32,7 +32,7 @@ from backend.protzilla.data_analysis.ptm_analysis import (
 )
 from backend.protzilla.data_analysis.ptm_quantification import flexiquant_lf
 from backend.protzilla.form import *
-from backend.protzilla.methods.data_preprocessing import TransformationLog
+from backend.protzilla.methods.data_preprocessing import TransformationLog, DataPreprocessingStep
 from backend.protzilla.steps import Plots, Step, StepManager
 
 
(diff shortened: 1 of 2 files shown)


backend/protzilla/methods/data_analysis.py
```diff
@@ -395,12 +628,57 @@ class DifferentialExpressionKruskalWallisOnIntensity(DataAnalysisStep):
...
+        ))
+
     calc_method = staticmethod(kruskal_wallis_test_on_intensity_data)
 
     def insert_dataframes(self, steps: StepManager, inputs) -> dict:
-        inputs["protein_df"] = steps.get_step_output(Step, "protein_df", inputs["protein_df"])
+        inputs["ptm_df"] = steps.get_step_output(Step, "ptm_df", inputs["ptm_df"])
         inputs["metadata_df"] = steps.metadata_df
-        inputs["log_base"] = steps.get_step_input(TransformationLog, "log_base")
         return inputs
```
AnnaPolensky, line 750: Just because I don't know the code base very well yet: Why are we using different input-files than before here?

AnnaPolensky (commented):
Things I've noticed by comparing the forms of Protzilla 2 and 3 (I've used the same input files):
- Some of the peptide analysis forms show exactly the same in Protzilla 3? + differ from Protzilla 2. E.g. "select peptides of protein" shows "ptms per sample" step?
- Protzilla 3 - Protein Quantification Plot: If I select Euclidean distance, it is still showing similarity as the next label, compared to Protzilla 2, where the next text box gets labelled with Euclidean distance in this case.
- Which steps did you skip deliberately? Was plot - precision call one of them? There is still no form defined. If you deliberately skip certain steps, perhaps you could mention them in your PR next time? (same applies to plot: receiver operating characteristic curve)
- modification quantification: FLEXIQUANT-LF does not load?/update to the same form as in Protzilla 2

Apart from that, I would say that the code looks good :)

backend/protzilla/methods/data_analysis.py
```diff
@@ -395,28 +693,117 @@ class DifferentialExpressionKruskalWallisOnIntensity(DataAnalysisStep):
...
+                    name="alpha",
+                    label="Error rate (alpha)",
+                    value=0.05,
+                    min=0,
+                    max=1,
+                    step=0.01,
+                    separatePrefix="\u03B1",
+                ),
+                DropdownField(name="grouping", label="Grouping from metadata"),
+                DropdownField(
```
3dot141592: Use `MultiSelectField` instead of `DropdownField` here, because the backend expects a list rather than a single string.

```suggestion
               MultiSelectField(
```

backend/protzilla/methods/data_analysis.py
```diff
@@ -395,28 +693,117 @@ class DifferentialExpressionKruskalWallisOnIntensity(DataAnalysisStep):
...
+        selected_groups_field.set_options(
+            form_helper.to_choices(run.steps.metadata_df[grouping].unique())
+        )
+
+    calc_method = staticmethod(kruskal_wallis_test_on_intensity_data)
+
+    def insert_dataframes(self, steps: StepManager, inputs) -> dict:
+        inputs["ptm_df"] = steps.get_step_output(Step, "ptm_df", inputs["ptm_df"])
+        inputs["metadata_df"] = steps.metadata_df
+        return inputs
```
3dot141592, line 750: **Use `protein_df` (plus `log_base`) instead of `ptm_df` in `insert_dataframes`** so the Kruskal-Wallis-on-intensity step uses the correct data source and passes the proper inputs to `kruskal_wallis_test_on_intensity_data`.

```suggestion
    def insert_dataframes(self, steps: StepManager, inputs) -> dict:
        inputs["protein_df"] = steps.get_step_output(
            Step, "protein_df", inputs["protein_df"]
        )
        inputs["metadata_df"] = steps.metadata_df
        inputs["log_base"] = steps.get_step_input(TransformationLog, "log_base")
        return inputs
```

backend/protzilla/methods/data_analysis.py
```diff
@@ -185,6 +202,49 @@ class DifferentialExpressionANOVA(DataAnalysisStep):
...
+                    name="alpha",
+                    label="Error rate (alpha)",
+                    value=0.05,
+                    min=0,
+                    max=1,
+                    step=0.01,
+                    separatePrefix="\u03B1",
+                ),
+                DropdownField(name="grouping", label="Grouping from metadata"),
+                DropdownField(
```
3dot141592: Use `MultiSelectField` instead of `DropdownField` here, because the backend expects a list rather than a single string.
```suggestion
                MultiSelectField(
```

backend/protzilla/methods/data_analysis.py
```diff
@@ -185,6 +202,49 @@ class DifferentialExpressionANOVA(DataAnalysisStep):
         "filtered_proteins",
     ]
 
+    def create_form(self):
+        return Form(
+            label="ANOVA",
+            input_fields=[
+                DropdownField(
+                    name="protein_df", label="Step to use protein intensities from"
```
3dot141592, line 210: The `protein_df` field in the form is currently not used: `insert_dataframes` ignores the user’s selection and always injects `steps.protein_df` as `intensity_df`.  

Fix: 

    def insert_dataframes(self, steps: StepManager, inputs) -> dict:
        inputs["log_base"] = steps.get_step_input(TransformationLog, "log_base")
        inputs["intensity_df"] = steps.get_step_output(
            Step, "protein_df", inputs["protein_df"]
        )
        inputs["metadata_df"] = steps.metadata_df
        return inputs
        
The same applies to other parts of the code.

backend/protzilla/methods/data_analysis.py
```diff
@@ -251,35 +311,45 @@ def create_form(self):
...
-            ))
+            group2_field.set_options(
+                list(
+                    reversed(
+                        form_helper.to_choices(run.steps.metadata_df[grouping].unique())
+                    )
+                )
+            )
 
     calc_method = staticmethod(t_test)
```
3dot141592, line 356: The same problem applies here with the fix:

```
def insert_dataframes(self, steps: StepManager, inputs) -> dict:
    inputs["log_base"] = steps.get_step_input(TransformationLog, "log_base")
    inputs["intensity_df"] = steps.get_step_output(
        Step, "protein_df", inputs["protein_df"]
    )
    inputs["metadata_df"] = steps.metadata_df
    return inputs
```

backend/protzilla/methods/data_analysis.py
3dot141592, line 1: In several places there are still UI fields whose values are not yet connected to the backend (i.e., they currently behave like no-op/dummy inputs), as well as a few remaining inconsistencies in the input flow. Since the PR description already states that input-flow TODOs will be handled separately, these points can be addressed in a follow-up issue.

backend/protzilla/methods/data_analysis.py
Elena-kal: In this step we are working with ptm which is also what the form is expecting as input. In other places we are working with proteins instead.

backend/protzilla/methods/data_analysis.py
Elena-kal: You are right. Is fixed now, thanks :))

backend/protzilla/methods/data_analysis.py
Elena-kal: I usually did not touch insert_dataframes, but you are right. Fixed it now.

backend/protzilla/methods/data_analysis.py
Elena-kal: You are right, is fixed.

PR #120, opened 2025-11-26
Add current_step_has_plot flag and improve plot placeholder message

### LINKED ISSUE #93: Misleading “No plot available” message before calculation
## Issue description
When entering a step that normally generates a plot, the interface immediately shows the message “No plot available for this step.” even if no calculation has been run yet. This is misleading because a plot will be available as soon as the user performs the calculation. The message correctly describes steps that never have a plot, but not steps where a plot depends on pending user input.

Alternative formulation:
- “Run calculation to generate plot.”
- "A plot will appear after calculation.”

## Acceptance Criteria
- [ ] Instead of “No plot available for this step.” there is an alternative formulation but only where it makes sense.
- [ ] The rest of the logic has not been touched and works like before.

3dot141592:
## Description
fixes #93 
Adjust plot placeholder text so that steps which can generate a plot no longer show a misleading “No plot available for this step.” message before the calculation is run.

## Changes
- **backend/main/views.py**
  - Add `current_step_has_plot` flag to `run_data` based on `run.current_step.plot_method`.
- **frontend/src/utils/protzilla-types.ts**
  - Extend `RunData` with `current_step_has_plot: boolean` and set it in `emptyRunData`.
- **frontend/src/components/app/run-screen/run-screen.tsx**
  - Use `runData.current_step_has_plot` to choose between:
    - `"Run calculation to generate plot."` (step can have a plot)
    - `"No plot available for this step."` (step never has a plot)

## Testing
- Open a step that can produce a plot but hasn’t been run yet → shows “Run calculation to generate plot.”
- Run the calculation on that step → plot appears as before.
- Open a step that never has a plot → still shows “No plot available for this step.”
- Navigate between steps with and without plots to ensure no errors and unchanged behavior apart from the placeholder text.

### DIFF
--- a/backend/main/views.py
+++ b/backend/main/views.py
@@ -392,7 +392,7 @@ def get_run_data(request):
             run_data["current_section"] = run.current_step.section
             run_data["current_step_index"] = run.steps.current_step_index
             run_data["memory_usage"] = get_memory_usage()
-            run_data["current_step_has_plot"] = bool(run.current_step.plot_method)
+            run_data["current_step_has_plot"] = True if run.current_step.plot_method is not None else False
         else:
             run_data["displayed_steps"] = []
             run_data["current_section"] = None
--- a/backend/main/views.py
+++ b/backend/main/views.py
@@ -392,11 +392,13 @@ def get_run_data(request):
             run_data["current_section"] = run.current_step.section
             run_data["current_step_index"] = run.steps.current_step_index
             run_data["memory_usage"] = get_memory_usage()
+            run_data["current_step_has_plot"] = bool(run.current_step.plot_method)
         else:
             run_data["displayed_steps"] = []
             run_data["current_section"] = None
             run_data["current_step"] = None
             run_data["memory_usage"] = get_memory_usage()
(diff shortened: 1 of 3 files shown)


backend/main/views.py
```diff
@@ -392,11 +392,13 @@ def get_run_data(request):
             run_data["current_section"] = run.current_step.section
             run_data["current_step_index"] = run.steps.current_step_index
             run_data["memory_usage"] = get_memory_usage()
+            run_data["current_step_has_plot"] = bool(run.current_step.plot_method)
```
Elena-kal: I think it would be clearer to read to simply write something like `True if run.current_step.plot_method is not None` instead of using bool() if I understand its intent correctly.

Elena-kal (approved):
Apart from this one comment it looks very reasonable.

backend/main/views.py
3dot141592: Thanks for the suggestion! I’ve replaced the bool(...) call with `True if run.current_step.plot_method is not None else False` to make it clearer.

jorisfu (approved):
mynd

PR #127, opened 2025-11-28
100 bug workflow import doesnt validate zip archive

### LINKED ISSUE #100: BUG: Workflow import doesn't validate zip archive
# What is the issue?
When importing a zip archive as a workflow template, the archive is always accepted if it is a valid archive. However, if there is no workflow.yaml present, it cannot be deleted from the UI with the error 'file workflow.yaml not found'

# How to reproduce the issue?
Upload any zip archive without a workflow.yaml in the root dir

Throw an error if the zip is missing files that are expected later on

tE3m:
## Description

fixes #100 

Before this fix, any kind of file (with any extension) could be uploaded as a workflow. But when deleting from the frontend, the code expected the file to be found at `WORKFLOWS_PATH/<name>.yaml`, so it couldn't ever be deleted.

## Changes

Now, the file picker only shows `.yaml` or `.yml` as accepted filetypes and the backend saves all files with the `.yaml` extension.

## Testing

1. Upload a workflow template
2. File is saved as .yaml in the correct folder
3. Run can be created from the + button
4. Workflow template can be deleted

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**

### DIFF
--- a/backend/protzilla/workflow.py
+++ b/backend/protzilla/workflow.py
@@ -7,7 +7,7 @@ def get_available_workflow_names() -> list[str]:
     return [
         file.stem
         for file in paths.WORKFLOWS_PATH.iterdir()
-        if not file.name.startswith(".") and not file.suffix == ".json"
+        if not file.name.startswith(".") and file.suffix == ".yaml"
     ]
 
 def delete_workflow_file(name: str) -> tuple[bool, str]:
--- a/backend/main/views.py
+++ b/backend/main/views.py
@@ -328,7 +328,7 @@ def import_workflow(request):
         workflow_file = settings.FILE_UPLOAD_TEMP_DIR / workflow
 
         if new_name == "":
-            copy2(str(workflow_file), str(WORKFLOWS_PATH / workflow))
+            copy2(str(workflow_file), str((WORKFLOWS_PATH / workflow).with_suffix(".yaml")))
         else:
             try:
                 copy2(str(workflow_file), str(WORKFLOWS_PATH / f"{new_name}.yaml"))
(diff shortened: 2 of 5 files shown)


NeleRiediger (commented):
Code changes look good to me. 
I just noticed while testing, that it's a little unintuitive that after successfully importing a workflow the pop-up doesn't close on its own. But this just might be another issue.

backend/protzilla/workflow.py
```diff
@@ -7,7 +7,7 @@ def get_available_workflow_names() -> list[str]:
     return [
         file.stem
         for file in paths.WORKFLOWS_PATH.iterdir()
-        if not file.name.startswith(".") and not file.suffix == ".json"
```
Elena-kal, line 10: Is there a reason why it was .json before and it is not checked for .json anywhere anymore? I thought the .yaml test was additional. If not maybe mention it in your pull request next time :))

Elena-kal (approved):
Apart of that one comment I think this is fine.

backend/protzilla/workflow.py
tE3m: Thank you for noting this, it had been done this way ever since [this commit](https://github.com/cschlaffner/PROTzilla2/commit/8559dfba9f3e4da06ce6e682e28decbdac9f3e87) in the old repo. Apparently, before workflows were described as `.yaml`, there were `.json` files present in the workflows' directory.
Before my changes, it was possible to upload any kind of file (e.g. a .zip archive) and have it be shown as a workflow, but now only ever actual workflow files are displayed.
I agree that I should have mentioned the reasons for this change

tE3m:
> Code changes look good to me. I just noticed while testing, that it's a little unintuitive that after successfully importing a workflow the pop-up doesn't close on its own. But this just might be another issue.

That's true, but I assume this was already the behaviour before and is out of scope of the issue at hand.

PR #128, opened 2025-11-29
92 bug import step loses state on revisit and blocks re import

### LINKED ISSUE #92: BUG: Import step loses state on revisit and blocks re-import
# What is the issue?
When navigating back to the file import step, the completed import is effectively undone. The UI still shows the previously selected file name, but clicking Import does nothing until the user reselects the same file. There is no error message visible in the frontend, but the console throws the following error:
```
[2025/11/08 - 12:49:38] [   ERROR] Internal Server Error: /api/calculate_step/
  File "/opt/anaconda3/envs/protzilla/lib/python3.11/site-packages/django/core/handlers/exception.py", line 56, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/anaconda3/envs/protzilla/lib/python3.11/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/elenakalbitzer/uni/WiSe2526/BP/PROTzilla/backend/main/views.py", line 480, in calculate_step
    return JsonResponse({"success": False, "message": calculation_data["messages"]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/anaconda3/envs/protzilla/lib/python3.11/site-packages/django/http/response.py", line 691, in __init__
    data = json.dumps(data, cls=encoder, **json_dumps_params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/anaconda3/envs/protzilla/lib/python3.11/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
          ^^^^^^^^^^^
  File "/opt/anaconda3/envs/protzilla/lib/python3.11/json/encoder.py", line 200, in encode
    chunks = self.iterencode(o, _one_shot=True)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/anaconda3/envs/protzilla/lib/python3.11/json/encoder.py", line 258, in iterencode
    return _iterencode(o, 0)
           ^^^^^^^^^^^^^^^^^
  File "/opt/anaconda3/envs/protzilla/lib/python3.11/site-packages/django/core/serializers/json.py", line 106, in default
    return super().default(o)
           ^^^^^^^^^^^^^^^^^^
  File "/opt/anaconda3/envs/protzilla/lib/python3.11/json/encoder.py", line 180, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type AssertionError is not JSON serializable
[2025/11/08 - 12:49:39] [   ERROR] "POST /api/calculate_step/ HTTP/1.1" 500 108114
```

# How to reproduce the issue?

- Open the import step and select a file.
- Click Import. The upload succeeds and a green check appears. Proceed to the next step.
- Navigate back to the import step.
- Click Import again without reselecting the file.

- The import step persists its state across navigation. Returning to the step shows the prior imported file as valid, with the green check still present.
- If the file must be reselected for technical reasons, the UI clearly indicates that the previous selection is no longer valid and disables Import until a new selection is made.

NeleRiediger:
## Description
fixes #92 
For the importing steps, when reverting back to a step the file input field shows "No file chosen" now instead of showing the name of the previously chosen file, which is no longer available and would cause an error when imported again.  

## Changes
In backend/protzilla/methods/importing.py the ImportingStep class was extended by the modify_form method (which is supposed to be overridden and called when clicking on a step). This method now changes the value of the input field of any ImportingStep based on its status so that the placeholder is shown instead of the previously selected file. 
For every ImportingStep that already had its own implementation of the modify_form method, a call to the super method was added (except for the ones that did not have an input field). 

## Testing
Upload a file in any importing step and click on "Import" and "Next" after the successful input. Now revert back to this step and see that the input field shows no file chosen, indicating correctly that a new file needs to be chosen before a new "Import". 

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**

### DIFF
--- a/backend/protzilla/methods/importing.py
+++ b/backend/protzilla/methods/importing.py
@@ -94,6 +94,10 @@ class MaxQuantImport(ImportingStep):
             ],
         )
 
+    def modify_form(self, form, run):
+        if run.steps.current_step.calculation_status == "complete":
+            form.input_fields[0].value = None
+
     calc_method = staticmethod(max_quant_import)
 
 
--- a/backend/protzilla/methods/importing.py
+++ b/backend/protzilla/methods/importing.py
@@ -65,7 +65,7 @@ class ImportingStep(Step):
     def index_of_file_input(self):
         """
         Returns the index of the FileInput that should be reset by modify_form. This method 
-        must be overriden if the FileInput is not index 0.    
+        must be overridden if the FileInput is not index 0.    
         """
         return 0
 
--- a/backend/protzilla/methods/importing.py
+++ b/backend/protzilla/methods/importing.py
@@ -57,6 +57,18 @@ class ImportingStep(Step):
     def insert_dataframes(self, steps: StepManager, inputs) -> dict:
         return inputs
 
+    def modify_form(self, form, run):
+        Step.modify_form(self, form, run)
+        if run.steps.current_step.calculation_status == "complete":
+            form.input_fields[self.index_of_file_input()].value = None
+
+    def index_of_file_input(self):
+        """
+        Returns the index of the FileInput that should be reset by modify_form. This method 
+        must be overriden if the FileInput is not index 0.    
+        """
+        return 0
+
 
 class MaxQuantImport(ImportingStep):
     display_name = "MaxQuant Protein Groups Import"
@@ -93,10 +105,6 @@ class MaxQuantImport(ImportingStep):
                 ),
             ],
         )
-
-    def modify_form(self, form, run):
-        if run.steps.current_step.calculation_status == "complete":
-            form.input_fields[0].value = None
 
     calc_method = staticmethod(max_quant_import)
 
@@ -326,6 +334,8 @@ class PeptideImport(ImportingStep):
         )
     
     def modify_form(self, form, run):
+        ImportingStep.modify_form(self, form, run)
+
         intensity_name_field = form["intensity_name"]
         map_to_uniprot_field = form["map_to_uniprot"]
 
@@ -368,6 +378,8 @@ class EvidenceImport(ImportingStep):
         )
     
     def modify_form(self, form, run):
+        ImportingStep.modify_form(self, form, run)
+
         map_to_uniprot_field = form["map_to_uniprot"]
 
         map_to_uniprot_field.value = run.steps.get_step_input(


backend/protzilla/methods/importing.py
```diff
@@ -57,6 +57,18 @@ def calc_method(self):
...
 
+    def modify_form(self, form, run):
+        Step.modify_form(self, form, run)
+        if run.steps.current_step.calculation_status == "complete":
+            form.input_fields[self.index_of_file_input()].value = None
+
+    def index_of_file_input(self):
+        """
+        Returns the index of the FileInput that should be reset by modify_form. This method 
+        must be overriden if the FileInput is not index 0.
```
3dot141592:
```suggestion
        must be overridden if the FileInput is not index 0.    
```

3dot141592 (approved):
Apart from that one typo, everything looks very clean.

backend/protzilla/methods/importing.py
```diff
@@ -57,6 +57,18 @@ def calc_method(self):
...
+        Step.modify_form(self, form, run)
+        if run.steps.current_step.calculation_status == "complete":
+            form.input_fields[self.index_of_file_input()].value = None
+
+    def index_of_file_input(self):
+        """
+        Returns the index of the FileInput that should be reset by modify_form. This method 
+        must be overriden if the FileInput is not index 0.    
+        """
+        return 0
```
Elena-kal, line 70: As of right now, we do not override this function anywhere and usually the index is at 0. Would it be cleaner to save index_of_file_input as a variable that can be overridden? Just because a function seems to be a bit overblown as we do not use it anywhere.

Elena-kal (changes requested):
The code apart from what I commented looks good. I was just wondering about the behavior we want. 
1. When I select a file and go to another step without importing it and then return to the importing step the file is still selected and there is no error when I import it then. Wouldn't it be nicer to have this behavior when I imported it already as well? 
2. Also why do we invalidate the step when we return to it? I know this was there before but wouldn't it make sense to not invalidate it since the file at that point is still imported. Users can and may still import different files but for that the step does not need to be invalidated right?

What I am saying is: The code looks good for what it is trying to accomplish but is this really the behavior we want?

PR #141, opened 2025-12-08
feat: add filtering based on SILAC ratios

tE3m:
## Description
fixes #134

adds the preprocessing step for filtering based on the amount of samples that a given protein has intensity records in.

## Changes

- rebased from #137  branch
- `backend/protzilla/data_preprocessing/filter_proteins.py`: add `by_silac_ratios` method
- `backend/protzilla/methods/data_preprocessing.py`: add `FilterProteinsBySilacRatios` step for the frontend
- `backend/tests/protzilla/data_preprocessing/test_filter_proteins.py`: add basic testing for the filtering

## Testing

- for whatever reason, when testing `by_silac_ratios` in the backend, everything works as expected. Using the [paper's dataset](https://ftp.pride.ebi.ac.uk/pride/data/archive/2020/04/PXD014997/SEARCH-IDENTIFICATIONS_REL_FREE-RELAPSE.7z) in an example workflow from the web UI always results in the same amount of proteins getting filtered, regardless of the amount enterd (the corresponding input would be 105). I could not reproduce this behaviour when debugging in pytest. Maybe it's just an odd error on my machine?

## PR checklist
**Development**
- [x] If necessary, I have updated the documentation (README, docstrings, etc.)
- [x] If necessary, I have created / updated tests.
 
**Mergeability**
- [x] main-branch has been merged into local branch to resolve conflicts
- [x] The tests and linter have passed AFTER local merge
- [x] The backend code has been formatted with `black`
- [x] The frontend code has been formatted with `pnpm format` and checked with `pnpm lint` (nothing in the frontend was touched)
 
**Code review**
- [x] I have self-reviewed my code.
- [ ] At least one other developer reviewed and approved the changes

PR #159, opened 2025-12-11
FIX: 157 silac import pulls wrong columns

### LINKED ISSUE #157: BUG: SILAC import pulls auxiliary columns
# What is the issue?
MaxQuant import included auxiliary SILAC-ratio columns (normalized/variability/count/type/peptides).

# How to reproduce the issue?
- Import a MaxQuant file with `Ratio H/L …`, `Ratio H/L normalized …`, `Ratio H/L variability …` etc.
- Take a closer look at the data. :(

- Only the selected ratio/intensity columns are imported; auxiliary columns are excluded.

3dot141592:
## Description
fixes #157  
Tighten MaxQuant ratio import so only the selected ratio columns are ingested and auxiliary columns (normalized, variability, count, type, peptides) are ignored for the non-normalized selection.

## Changes
- Update `backend/protzilla/importing/ms_data_import.py` to build a stricter regex for ratio imports, excluding normalized/auxiliary columns unless explicitly chosen.
- This Branch is based on #141; So please ignore all other changes except `backend/protzilla/importing/ms_data_import.py`

## Testing
- `python3 -m pytest backend/tests/protzilla/importing/test_ms_data_import.py::test_max_quant_import_different_intensity_names`
- Looking at the data

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**

### DIFF
--- a/backend/protzilla/importing/ms_data_import.py
+++ b/backend/protzilla/importing/ms_data_import.py
@@ -36,7 +36,9 @@ def max_quant_import(
         ):
             base_pattern = rf"^{re.escape(intensity_name)}\s(?!normalized\b)(?!.*\b{disallowed_suffixes}\b).*$"
         else:
-            base_pattern = rf"^{re.escape(intensity_name)}\s(?!.*\b{disallowed_suffixes}\b).*$"
+            base_pattern = (
+                rf"^{re.escape(intensity_name)}\s(?!.*\b{disallowed_suffixes}\b).*$"
+            )
 
         intensity_df = df.filter(regex=base_pattern, axis=1)
 
--- a/backend/protzilla/importing/ms_data_import.py
+++ b/backend/protzilla/importing/ms_data_import.py
@@ -29,9 +29,16 @@ def max_quant_import(
             keep_default_na=True,
         )
         protein_groups = df["Majority protein IDs"]
-        # filter out normalized so "Ratio H/L normalized" is not filtered for "Ratio H/L" - same for L/H
-        intensity_df = df.filter(regex=f"^{intensity_name} (?!normalized)", axis=1)
-        intensity_df = intensity_df.filter(regex=r"^(?!.*peptides).*$", axis=1)
+        disallowed_suffixes = r"(variability|count|type|peptides)"
+        if intensity_name in (
+            IntensityType.RATIO_HL.value,
+            IntensityType.RATIO_LH.value,
+        ):
+            base_pattern = rf"^{re.escape(intensity_name)}\s(?!normalized\b)(?!.*\b{disallowed_suffixes}\b).*$"
+        else:
+            base_pattern = rf"^{re.escape(intensity_name)}\s(?!.*\b{disallowed_suffixes}\b).*$"
+
+        intensity_df = df.filter(regex=base_pattern, axis=1)
 
         if intensity_df.empty:
             msg = f"{intensity_name} was not found in the provided file, please use another intensity and try again or verify your file."
--- a/backend/protzilla/importing/peptide_import.py
+++ b/backend/protzilla/importing/peptide_import.py
@@ -1,5 +1,6 @@
 import logging
 from pathlib import Path
+import re
 
 import pandas as pd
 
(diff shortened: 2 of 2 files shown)


backend/protzilla/importing/ms_data_import.py
```diff
@@ -29,9 +29,18 @@ def max_quant_import(
             keep_default_na=True,
         )
         protein_groups = df["Majority protein IDs"]
-        # filter out normalized so "Ratio H/L normalized" is not filtered for "Ratio H/L" - same for L/H
-        intensity_df = df.filter(regex=f"^{intensity_name} (?!normalized)", axis=1)
-        intensity_df = intensity_df.filter(regex=r"^(?!.*peptides).*$", axis=1)
+        disallowed_suffixes = r"(variability|count|type|peptides)"
```
AnnaPolensky, line 32: Are there suffixes which we allow? Otherwise, we could just disallow all suffixes? Although, it would also make sense to me to specifically state what exactly we want to disallow, so might be fine?

backend/protzilla/importing/ms_data_import.py
```diff
@@ -29,9 +29,18 @@ def max_quant_import(
...
+        if intensity_name in (
+            IntensityType.RATIO_HL.value,
+            IntensityType.RATIO_LH.value,
+        ):
+            base_pattern = rf"^{re.escape(intensity_name)}\s(?!normalized\b)(?!.*\b{disallowed_suffixes}\b).*$"
```
AnnaPolensky, line 37: 1) Could we maybe just add "normalized" to disallowed_suffixes?

2) And do we need the distinction between Ratio and non-ratio intensity types? Because the regex does look exactly the same apart from the (?!nomalized\b) which shouldn't cause harm to other intensity types?

3) I'm just not that familiar with regex, why do we need re.escape?

AnnaPolensky (commented):
If this is a problem for the protein import, it might also be a problem for the peptide import? So, should we fix the regex for the peptide import as well?

Elena-kal (commented):
All in all the code looks fine. However there are still the following points to adress:

1. The same thing should be done in [peptide_import.py](backend/protzilla/importing/peptide_import.py) like Anna said. There we still do this:
```
if "Sample" not in df.columns:
        id_df = df[id_columns]
        # filter out normalized so "Ratio H/L normalized" is not filtered for "Ratio H/L" - same for L/H
        intensity_df = df.filter(regex=f"^{intensity_name} (?!normalized)", axis=1)
```
2. Also why is this branch based on #141 ? Is there a reason why it is not branched from dev? And shouldn't we merge this into #141 because there are still changes made on #141 right? Or at least rebase and wait for the branch to merge into dev.

backend/protzilla/importing/ms_data_import.py
3dot141592: On `disallowed_suffixes`: We still need the sample name as the suffix, so we can’t block everything unfortunetly. And the sample names often differs as well so we cant filter for that eather (e.g CTR#, P#, ...). The intent is to drop known wrong fields (variability/count/type/peptides); if there are more wrong columns we should add them, but to just disallow all of them would also remove the actual sample columns.

backend/protzilla/importing/ms_data_import.py
3dot141592: `Normalized` can’t go into `disallowed_suffixes` because we still need to import normalized ratios when the user selects them; the (?!normalized\b) applies only to the non-normalized choice.
• The ratio vs. non-ratio branch is there because only ratios have the normalized "twin";
• `re.escape` makes sure characters like the slash in Ratio H/L are treated as literals, not regex operators.

3dot141592:
First of all: Thanks, @AnnaPolensky and @Elena-kal , for this quick unprepared review. 

And good catch – the regex fix should go into peptide_import.py as well. Thanks for pointing that out.

3dot141592:
>     2. Also why is this branch based on #141 ? Is there a reason why it is not branched from dev? And shouldn't we merge this into #141 because there are still changes made on #141 right? Or at least rebase and wait for the branch to merge into dev.

This branch is based on #141, as the issue came up while I was reviewing that pull request. There shouldn’t be any merge conflicts afaik, so we should be able to merge this PR once #141 has been merged.

tE3m:
there have been changes to #141, so don't forget to rebase before merging

backend/protzilla/importing/ms_data_import.py
AnnaPolensky: Thank you for the explanation. That makes a lot of sense!

backend/protzilla/importing/ms_data_import.py
AnnaPolensky: From my understanding the normalized will be part of the intensity_name if we are looking for normalized columns. Therefore, I thought that it should work if we always exclude the normalized because if we aren't using ratios it should not cause any harm to also filter out normalized and if normalized is part of the intensity_name it should be fine to filter out columns that are containing normalized twice?
I believe that your version works, I was just wondering whether we could do something simpler.
Thanks for the explanation of re.escape, that makes a lot of sense.

3dot141592:
the last force push was a rebase for #141 :)

Issue #132, opened 2025-12-02
AML Paper: Adjust Analysis Steps to Work With Ratios


jorisfu:
## TODO description
The following analysis steps need to work with ratio values:

- t tests
- hierarchical clustering
- GO analysis

PR #170, opened 2026-01-04
implement SILAC t-test

3dot141592:
## Description
fixes one part of #132 
- Make the t-test tolerate SILAC ratios with NaNs and still show its tables.

## Changes
- differential_expression_t_test.py: drop NaNs, require ≥2 values/group.
- test_differential_expression.py: add SILAC ratio regression test (p ≈ 0.0513, log2FC ≈ -0.44).
- views.py: include all *_df outputs in the “Tables” tab.

## Testing
- PYTHONPATH=backend pytest backend/tests/protzilla/data_analysis/test_differential_expression.py -k silac --maxfail=1
- Frontend: run t-test and check tables.
- You can create your own metadata or use the one i created: 
[metadata.csv](https://github.com/user-attachments/files/24424576/metadata.csv)

## PR checklist
**Development**

**Mergeability**

**Code review**

## Known Issues
- If a Table is empty the whole frontend crashes but this was most likely not done through this PR.

### DIFF
--- a/backend/main/views.py
+++ b/backend/main/views.py
@@ -633,8 +633,8 @@ def get_step_table(request):
 
         if run.current_step is not None:
             for key, value in run.current_outputs:
-                if key in dataframes:
-                    data = value
+                if isinstance(value, pd.DataFrame):
+                    data = value.copy()
                     data["id"] = data.index
                     cleaned_data = data.replace(np.nan, None)
                     json_data.append(
--- a/backend/protzilla/data_analysis/differential_expression_t_test.py
+++ b/backend/protzilla/data_analysis/differential_expression_t_test.py
@@ -101,6 +101,14 @@ def t_test(
         protein_df = intensity_df[intensity_df["Protein ID"] == protein]
         group1_intensities = protein_df[protein_df[grouping] == group1][intensity_name]
         group2_intensities = protein_df[protein_df[grouping] == group2][intensity_name]
+
+        group1_intensities = group1_intensities.dropna()
+        group2_intensities = group2_intensities.dropna()
+        if len(group1_intensities) < 2 or len(group2_intensities) < 2:
+            if not exists_message(messages, INVALID_PROTEINGROUP_DATA_MSG):
+                messages.append(INVALID_PROTEINGROUP_DATA_MSG)
+            continue
+
         t, p = stats.ttest_ind(
             group1_intensities,
             group2_intensities,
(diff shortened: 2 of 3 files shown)


backend/main/views.py
```diff
@@ -633,8 +633,8 @@ def get_step_table(request):
 
         if run.current_step is not None:
             for key, value in run.current_outputs:
-                if key in dataframes:
-                    data = value
+                if isinstance(value, pd.DataFrame):
+                    data = value.copy()
```
AnnaPolensky, line 637: Why is the copy necessary now that we are working with ratios?

backend/protzilla/data_analysis/differential_expression_t_test.py
```diff
@@ -101,6 +101,14 @@ def t_test(
         protein_df = intensity_df[intensity_df["Protein ID"] == protein]
         group1_intensities = protein_df[protein_df[grouping] == group1][intensity_name]
         group2_intensities = protein_df[protein_df[grouping] == group2][intensity_name]
+
+        group1_intensities = group1_intensities.dropna()
+        group2_intensities = group2_intensities.dropna()
+        if len(group1_intensities) < 2 or len(group2_intensities) < 2:
```
AnnaPolensky, line 107: Why do we choose 2 here? Maybe we could use a variable that contains 2 here instead, in order to avoid magic numbers?

Elena-kal (commented):
I thought this PR was also supposed to check whether it works the way it is done in the paper (I thought it was something about checking variance not expectation but I do not quite remember and I could not find it in our protocols. But we definetly talked about it during sprint planning and that was why we split the issue in two). 

I do not know whether you checked the implementation of the t-test an it is exactly what we wanted and that is why you did not change anything. If so maybe mention that in the PR next time so everyone knows that the issue is actually solved.

backend/protzilla/data_analysis/differential_expression_t_test.py
3dot141592: Mathematically speaking, the sample variance, which we will later need for the t-test, is not defined for sets with fewer than two elements. Therefore, I have excluded these cases.

Happy to lift the 2 into a named constant to avoid the magic number if you prefer, but this number will never change, so it could stay like this imo.

backend/main/views.py
3dot141592: copy() prevents us from mutating the step’s stored DataFrame when we add the helper id column for the frontend. Without the copy, every time we fetch the table, it would push that id column back into the run outputs, which is an unintended side effect and bloats our ram usage.

  This might be one of multiple sources of the every increasing ram usage of PROTZilla.

PR #171, opened 2026-01-06
132 aml paper adjust analysis steps for ratios: hierarchical clustering

3dot141592:
## Description
Fixes the second part of #132 (Fixes hierarchical clustering).

## Changes
- Changed the clustering scoring values over to the real sklearn metric IDs.
- Populate the label / positive-label fields from the metadata.
- Use metadata_df everywhere instead of sample_group_df (old behaviour i think).

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**


backend/protzilla/methods/data_analysis.py
```diff
@@ -1327,9 +1342,24 @@ def create_form(self):
             ],
         )
 
+    def modify_form(self, form, run):
```
NeleRiediger, line 1460: It doesn't feel clean to add the exact same block of code several times in one file. This kind of redundant code seems to happen a lot in this file: the code you added seems to be exactly the same three times and in the code block where you exchanged "sample_group_df" for "metadata_df" seems to appear at least five times. I feel like we should find a way to resolve these redundancies by pulling the code up one layer or something?

NeleRiediger (changes requested):
The new code itself looks fine, I just think the changed file needs some refactoring so the new code isn't copy pasted several times in it.

backend/protzilla/methods/data_analysis.py
3dot141592: You’re right about the duplication, but I don’t think refactoring it is in scope for this issue/PR. Let’s keep this PR focused for now, and handle the cleanup in a follow-up - feel free to open a separate issue for it. I think we’ll make a lot of this cleaner anyway when we refactor the workflow.

backend/protzilla/methods/data_analysis.py
```diff
@@ -115,14 +115,14 @@ class ClusteringCriterion(Enum):
...
-    rand_score = "Rand Score"
-    v_measure_score = "V Measure Score"
+    adjusted_rand_score = "adjusted_rand_score"
+    completeness_score = "completeness_score"
+    fowlkes_mallows_score = "fowlkes_mallows_score"
+    homogeneity_score = "homogeneity_score"
+    mutual_info_score = "mutual_info_score"
+    normalized_mutual_info_score = "normalized_mutual_info_score"
+    rand_score = "rand_score"
+    v_measure_score = "v_measure_score"
```
tE3m, line 125: This is ok for the moment since it works, but I don't think it's nice for users to have to read the internal names. Adjusted Rand Score reads much nicer than adjusted_rand_score.
From what I can see, this change could also be executed in `classification_helper.py`, but without losing the human-friendly naming.

tE3m (commented):
Code looks good, but I'd like to hear your opinion regarding the comment. I also failed at testing the feature, a description of what to import/do would have been nice. I got this error and don't know what to do/change:

```
An error occured while validating inputs or outputs: For multi-metric scoring, the parameter refit must be set to a scorer key or a callable to refit an estimator with the best parameter setting on the whole data and make the best_* attributes available for that metric. If this is not needed, refit should be set to False explicitly. 'homogeneity_score' was passed. Please check your parameters.

  File "/home/prot/zilla/backend/protzilla/steps.py", line 125, in calculate
    calc_output = self.calc_method(**self.calculation_input)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/prot/zilla/backend/protzilla/data_analysis/clustering.py", line 321, in hierarchical_agglomerative_clustering
    model, model_evaluation_df = perform_clustering(
                                 ^^^^^^^^^^^^^^^^^^^
  File "/home/prot/zilla/backend/protzilla/data_analysis/clustering.py", line 370, in perform_clustering
    model.fit(input_df, labels_df)
  File "/usr/local/lib/python3.11/site-packages/sklearn/base.py", line 1365, in wrapper
    return fit_method(estimator, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/sklearn/model_selection/_search.py", line 1060, in fit
    self._check_refit_for_multimetric(first_test_score)
  File "/usr/local/lib/python3.11/site-packages/sklearn/model_selection/_search.py", line 819, in _check_refit_for_multimetric
    raise ValueError(multimetric_refit_msg)
ValueError: For multi-metric scoring, the parameter refit must be set to a scorer key or a callable to refit an estimator with the best parameter setting on the whole data and make the best_* attributes available for that metric. If this is not needed, refit should be set to False explicitly. 'homogeneity_score' was passed.

```

backend/protzilla/methods/data_analysis.py
NeleRiediger: I feel like we shouldn't write "bad" code to fix something quickly and then open a follow-up issue to refactor it, as this builds technical debt. Why not do it right the first time? @tE3m do you have an opinion on that? 
Maybe that's also something to generally discuss with the whole team to maybe have a consensus for such cases in the future.

PR #175, opened 2026-01-08
Improved runner tests (and bug fixes)

hendraet:
## Description

The purpose of this PR was mainly to test if other data (MSFragger, DIA-NN) are properly working with steps down the line - so an integration test. The runner test seemed like a good way to test it, so I added an appropriate test.
Turns out the other runner tests were actually not working properly (despite passing), so I fixed them as well. 

While fixing the tests, I ran into an unhandled edge case for differential expressions, so I decided to handle it and add appropriate tests.

## Testing

Mainly just run the tests in `test_runner.py` and `test_differential_expression.py`, but you can also manually load some MSFragger data into a standard workflow where the protein import step has been changed accordingly, or you just import the `MSFragger_Standard.yaml` as a new workflow. 

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**


backend/protzilla/data_analysis/differential_expression_mann_whitney.py
```diff
@@ -236,6 +236,41 @@ def mann_whitney_test_on_columns(
...
+                    "u_statistic",
+                ]
+            ),
+            significant_columns_df=pd.DataFrame(
+                columns=[
+                    columns_name,
+                    "corrected_p_value",
+                    "log2_fold_change",
+                    "u_statistic",
+                ]
```
AnnaPolensky: nit: maybe define columns once and use in both places?

backend/protzilla/methods/data_integration.py
```diff
@@ -831,7 +831,7 @@ class PlotGOEnrichmentBarPlot(PlotStep):
     operation = "plot"
     method_description = "Creates a bar plot from GO enrichment data"
 
-    output_keys = ["plots"]
+    output_keys = []
```
AnnaPolensky, line 834: why can we just remove "plots" from the output_keys in all these places?

backend/tests/protzilla/test_runner.py
```diff
@@ -56,6 +65,86 @@ def mock_current_parameters(*args, **kwargs):
...
+    """Find a step by class name and set multiple field values."""
+    step_idx = find_step_by_class_name(runner, class_name)
+    for field_name, value in field_values.items():
+        set_step_field_value(runner, step_idx, field_name, value)
+    return step_idx
+
+
+def prepare_standard_workflow_runner(runner: Runner):
+    """
+    The standard worfklow misses leaves some of the configurable fields blank because it is a general purpose workflow
```
AnnaPolensky: nit: using "misses" and "leaves" both does not make sense grammatically

backend/tests/protzilla/test_runner.py
```diff
@@ -281,16 +370,87 @@ def test_integration_runner(
             "meta_data_path": f"{TEST_METADATA_PATH}/{metadata_file_path}",
             "peptides_path": None,
             "run_name": f"{name}",
-            "df_mode": "disk",
+            "df_mode": "memory",
```
AnnaPolensky, line 385: why does it make sense to change this?

backend/tests/protzilla/test_runner.py
```diff
@@ -281,16 +370,87 @@ def test_integration_runner(
...
+    monkeypatch,
+):
+    name = tests_folder_name + "/test_runner_integration_" + random_string()
+
+    standard_workflow_file = TEST_WORKFLOWS_PATH / f"{mock_workflow}.yaml"
+    shutil.copy(standard_workflow_file, tmp_workflow_dir)
+    with mock.patch.object(
+        disk_operator.paths, "WORKFLOWS_PATH", tmp_workflow_dir.resolve()
+    ):
+        print("ADBLHBSFHLB: ", f"{TEST_MSDATA_PATH}/{ms_data_file_path}")
```
AnnaPolensky: What does "ADBLHBSFHLB" stand for? Do we really want to have a print statement in our code?

backend/tests/protzilla/test_runner.py
```diff
@@ -281,16 +370,87 @@ def test_integration_runner(
...
+        runner = Runner(
+            **{
```
AnnaPolensky: From my understanding of kwargs, we are creating a dict here just to immediately unpack it, is that right?

backend/tests/protzilla/test_runner.py
```diff
@@ -304,11 +464,22 @@ def test_integration_runner_no_plots(
...
     monkeypatch.setattr(runner.run, "_run_write", mock_write)
     runner.compute_workflow()
+    assert all(
+        step.calculation_status == "complete" for step in runner.run.steps.all_steps
+    )
+    assert runner.run.steps.all_steps[-1] == runner.run.current_step
+    assert (
+        all(step.finished for step in runner.run.steps.all_steps)
+        and not runner.run.current_step.messages
+        and "messages" not in runner.run.current_step.output
```
AnnaPolensky: I think this is the third time I'm seeing exactly this code. Would it make sense to extract this condition to some function?

AnnaPolensky (commented):
Tests also ran successfully locally on my machine. Code looks pretty good.

backend/protzilla/methods/data_integration.py
hendraet: This was one of the bugs that gave me major headaches and the reason why these steps were never marked as `finished`. (During output validation, it is checked if outputs contain a `plots` key, which is not the case because plots are saved separately in the `plots` attribute) 
I'm honestly not sure where this came from, but I guess that it is legacy code before plots were saved in a dedicated attribute. I couldn't find steps outside of data integration that still have this flag.

backend/tests/protzilla/test_runner.py
hendraet: I like that you are asking all the right questions :) 
This was another important point that initially broke the test runners. Using disk mode is only a good idea if you are **not** mocking the write method a couple lines later :sweat_smile: 
This lead to dataframes not being loaded and thus the first preprocessing step immediatly threw an error.

backend/tests/protzilla/test_runner.py
hendraet: Whoopsie :see_no_evil:

NeleRiediger (approved):

Issue #194, opened 2026-01-17
Display the failed rows after a cross linking import


NeleRiediger:
## Issue description
After a cross linking import the user gets a message (success or warning) on how many rows of the table were successfully imported and how many rows could not be processed as they contain any kind of data we cannot process (e.g. experimental names). The user should also get information on which rows couldn't be imported and why a respective row couldn't be imported. 
! We still need to make a decision on how to display this table of failed rows ! 
(ideas: expendable warning message, showing the failed rows table full screen on click
or having the failed rows as an output data frame besides the successfully imported data) 

## Acceptance Criteria
- [ ]

NeleRiediger:
The final decision on how to implement this, is to show the failed_df along with the good_df in the tables tab. Besides that the warning right after importing should clearly show the failed protein designations along with their respective errors.

Issue #193, opened 2026-01-17
Add an isoforms column to our cross linking data frame


NeleRiediger:
## Issue description
Since we might need the isoform information when working with cross links and a specific spatial protein structure, but didn't consider this when specifying our desired data frame for cross linking data, this column needs to be added. Moreover it needs to be filled with the respective data. 
CSM file import: those file only store gene name (no protein ids containing isoform information). However the UniProt lookup fetching the respective protein id to a gene name already also get all respective isoforms. This information just needs to be stored in the new column now instead of not being stored at all. 
Protein Discoverer XLinkX file import: those files already store protein ids and some of them are isoforms. Initially we remove the isoform information from those ids to be able to process them. So those isoforms just need to be stored in the new column before removing the information for the protein id column. (Since not all ids contain isoform information, a decision on what to do with those rows needs to be made) 

## Acceptance Criteria
- [ ] The documentation of our data frame has an isoform column with all its necessary information 
- [ ] The data frame has an isoform column 
- [ ] No matter what file type is imported, the isoform column is filled with isoform information (as we specified)

PR #195, opened 2026-01-17
Add importing step for crosslinking data

### LINKED ISSUE #192: Add Cross Linking Import
## Issue description
A new step for the import of cross linking data should be created. It should be able to import CSM (.csv) cross linking files as well as Protein Discoverer XLinkX (.xlsx) files. No matter which of those two file types was imported, after the import the relevant data from this file is stored in our standardized data frame as documented in our slack channel. Particularly the designation of the proteins: CSM files only contain the gene name of their involved proteins (e.g. RAD50) whereas Protein Discoverer XLinkX files only contain the protein id (e.g. P62195) of their involved proteins. Since we want a standardized data frame we choose to store both designation variants for every file. So in each import to get the respective missing protein designations we use UniProt lookups via their API to fetch those designations based on the version we already have. Data rows of the imported file containing any kind of data we cannot process (e.g. experimental names) should be kept out of the final data frame, so they don't prevent the whole file from getting imported. This information should be displayed to the user. 

## Acceptance Criteria
- [ ] There is a new step for the cross linking import 
- [ ] This step imports CSM (.csv) cross linking files as well as Protein Discoverer XLinkX (.xlsx) files successfully 
- [ ] The output table has all the columns as specified in our documentation of the data frame. 
- [ ] All columns of the output table contain seemingly valid information. 
- [ ] Particularly both versions of designation of the proteins exists and seem valid. 
- [ ] After the import is completed the user gets information, if every single row of data could be processed or if there were rows containing data that prevented a successful import of this row (saying how many rows are affected, which und for what cause the respective row was rejected)

NeleRiediger:
## Description
fixes #192 

There is a new step for the import of cross linking data. It is able to import CSM (.csv) cross linking files as well as Protein Discoverer XLinkX (.xlsx) files. No matter which of those two file types was imported, after the import the relevant data from this file is stored in our standardized data frame as documented in our slack channel. 
Particularly the designation of the proteins: CSM files only contain the gene name of their involved proteins (e.g. RAD50) whereas Protein Discoverer XLinkX files only contain the protein id (e.g. P62195) of their involved proteins. Since we want a standardized data frame we choose to store both designation variants for every file. So in each import to get the respective missing protein designations we use UniProt lookups via their API to fetch those designations based on the version we already have. 
Data rows of the imported file containing any kind of data we cannot process (e.g. experimental names) are kept out of the final data frame, so they don't prevent the whole file from getting imported. This information is displayed to the user via Warnings after the import. (Right now a table of the rows that failed to import is stored in a file "failed_df.csv" that is created after a successful import. Since I thought about different ways to display this information in Protzilla including a possible change to warning messages (making them expand on click), I felt like this should be an extra issue #194.)
(Moreover there is a second follow up issue, since this one was already big enough: #193) 

## Changes

- backend/protzilla/methods/importing.py ➞ creating a new step 
- backend/protzilla/all_steps.py ➞ adding the new step 
- backend/protzilla/importing/cross_linking_import.py ➞ most of the changes, including reading the data from a file, preprocessing it concerning the standardization and Unirot lookups and then storing it in the dataframe created 

## Testing

Test the new step for the cross linking import:
- import a CSM (.csv) cross linking file as well as Protein Discoverer XLinkX (.xlsx) file (to see if both is successful) 
- look at the output table after a successful import to see if it contains all columns as specified in our documentation of the data frame (in slack)
- look if all the data in the output table is seemingly valid information
- look at the failed_df.csv that was created, to see if any row in there might be rejected even though it shouldn't  

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**


backend/protzilla/importing/cross_linking_import.py
```diff
@@ -0,0 +1,512 @@
...
+    Extracts unique values from two DataFrame columns and returns them as a set.
+
+    Parameters:
+        df (pd.DataFrame): Input DataFrame
+        column (str): Column name
+
+    Returns:
+        set: Unique values from the columns
+    """
+    return set(
```
jorisfu: This does not follow the [reST docstring pattern](https://realpython.com/how-to-write-docstrings-in-python/#restructuredtext-docstrings) we use throughout the project. Same applies for the other functions

backend/protzilla/importing/cross_linking_import.py
```diff
@@ -0,0 +1,512 @@
...
+        df[[column + "1", column + "2"]].stack().dropna().astype(str).str.strip()
+    )
+
+
+def validate_data_before_lookup(
+    data_for_lookup: set, is_valid_function, error_code: str
```
jorisfu: `is_valid_function` sounds like a boolean value, use `validator_function` or similar instead

backend/protzilla/importing/cross_linking_import.py
```diff
@@ -0,0 +1,512 @@
...
+):
+    """
+    Splits input values into valid and invalid ones.
+
+    Invalid values are directly written to results with the given error code.
+
+    Returns:
+        valid_data (set)
+        results (dict): value -> (False, None, error_code)
```
jorisfu: This format seems kinda out of the blue for me, why do we need `False, None` here if they are always the same value for every bad data?

backend/protzilla/importing/cross_linking_import.py
```diff
@@ -0,0 +1,512 @@
...
+            if g in valid_input:
+                if "-" in protein_id:
+                    output[g]["list_of_protein_isoforms"].append(protein_id)
+                else:
+                    output[g]["protein_ids"].append(protein_id)
+            elif isFallback:
+                if "-" in protein_id:
+                    output[valid_input]["list_of_protein_isoforms"].append(protein_id)
+                else:
+                    output[valid_input]["protein_ids"].append(protein_id)
```
jorisfu: `valid_input` is an iterable **and** a key in `output`? 🤨

backend/protzilla/importing/cross_linking_import.py
```diff
@@ -0,0 +1,512 @@
...
+    Returns:
+        dict: Mapping gene_name -> (success, data, error)
+            success (bool): True if lookup for this gene_name succeeded, False otherwise
+            data (dict or None): {
+                "protein_ids" (list of str): all protein IDs without any isoform information,
+                "list_of_protein_isoforms" (list of str): all isomform IDs
+                } if success else None
+            error (str or None): error code/message if failed, else None
+    """
+    # Filter decoy Proteins, because we cannot process them decently?
```
jorisfu: question mark?

backend/protzilla/importing/cross_linking_import.py
```diff
@@ -0,0 +1,512 @@
...
+    data_for_lookup: set,
+    field_of_existing_data: str,
+    *,
+    extra_query: str | None = None,
+    response_format: str,
+    fields: str,
+    include_isoforms: bool = False,
+):
+    """
+    Builds UniProt search URL and params.
```
jorisfu: some more docs would be nice here (params)

backend/protzilla/importing/cross_linking_import.py
```diff
@@ -0,0 +1,512 @@
...
+        "fields": fields,
+    }
+
+    if include_isoforms:
+        params["includeIsoform"] = "true"
+
+    return uniprot_search_url, params
+
+
+def execute_uniprot_request(url, params, valid_data, results):
```
jorisfu: missing docstring

backend/protzilla/importing/cross_linking_import.py
```diff
@@ -0,0 +1,512 @@
...
+        error = f"HTTP_{e.response.status_code}"
+    except requests.exceptions.RequestException:
+        error = "REQUEST_ERROR"
+
+    for data in valid_data:
+        results[data] = (False, None, error)
+    return None
+
+
+def process_uniprot_response_containing_gene_names(response, results):
```
jorisfu: add docstring

backend/protzilla/importing/cross_linking_import.py
```diff
@@ -0,0 +1,512 @@
...
+        output = entry.get("genes", [{}])
+        gene_name = output[0].get("geneName", {}).get("value") if output else None
+
+        if gene_name:
+            results[protein_id] = (True, gene_name, None)
+        else:
+            results[protein_id] = (False, None, "NO_GENE_NAME_FOUND")
+
+
+def process_uniprot_response_containing_protein_ids(
```
jorisfu: add docstring

backend/protzilla/importing/cross_linking_import.py
```diff
@@ -0,0 +1,512 @@
...
+                    output[g]["protein_ids"].append(protein_id)
+            elif isFallback:
+                if "-" in protein_id:
+                    output[valid_input]["list_of_protein_isoforms"].append(protein_id)
+                else:
+                    output[valid_input]["protein_ids"].append(protein_id)
+    return output
+
+
+def fallback_single_lookup(query: str, query_type: str, results):
```
jorisfu: add docstring

jorisfu (commented):
Good code style, generally like it. Just please reformat the docstrings (else people with IDEs are not gonna profit from them) and add some more docstrings for functions that aren't immediately obvious (especially when using custom tuple formats throughout multiple functions – I'd prefer a struct generally, but it's OK here).

Also I'd say it's fine that we have both csv and xlsx imports in one steps. The code is structured well enough that this doesn't cause any major chaos IMO.

jorisfu:
Also please add tests

jorisfu:
Note: Importing `20230728_SFB_BS3_Gel_MRN_ATM_S3_Xi1.8.7_CSM_xiFDR2.3.5.csv` we were given does not work:

An error occurred while reading the file: KeyError "Only a column name can be used for the key in a dtype mappings argument. 'Crosslinker' not found in columns.". Please provide a valid cross linking file.

backend/protzilla/importing/cross_linking_import.py
```diff
@@ -0,0 +1,512 @@
...
+        gene_name = output[0].get("geneName", {}).get("value") if output else None
+
+        if gene_name:
+            results[protein_id] = (True, gene_name, None)
+        else:
+            results[protein_id] = (False, None, "NO_GENE_NAME_FOUND")
+
+
+def process_uniprot_response_containing_protein_ids(
+    response, valid_input, isFallback: bool
```
Elena-kal:
```suggestion
    response, valid_input, is_fallback: bool
```

backend/protzilla/importing/cross_linking_import.py
```diff
@@ -0,0 +1,512 @@
...
+    header = lines[0].split("\t")
+    protein_id_idx = header.index("Entry")
+    gene_name_idx = header.index("Gene Names (primary)")
+
+    for line in lines[1:]:
+        parts = line.split("\t")
+        protein_id = parts[protein_id_idx]
+        output_gene_names = parts[gene_name_idx].split()
+
+        for g in output_gene_names:
```
Elena-kal: nitpick but maybe gene instead of g? :))

backend/protzilla/importing/cross_linking_import.py
```diff
@@ -0,0 +1,512 @@
...
+
+    response = execute_uniprot_request(url, params, valid_gene_names, results)
+    if response is None:
+        return results
+
+    output = process_uniprot_response_containing_protein_ids(
+        response, valid_gene_names, False
+    )
+
+    for gn in valid_gene_names:
```
Elena-kal: Again nitpick but maybe gene instead of gn?

backend/protzilla/importing/cross_linking_import.py
```diff
@@ -0,0 +1,512 @@
...
+    value_extractor=lambda x: x,
+):
+    """
+    Iterates over a DataFrame and adds the missing protein designations to the dataframe using
+    precomputed lookup results. (either protein ids or gene names are included in the imported
+    data and the other is added to the data frame here)
+
+    Parameters:
+        df (pd.DataFrame)
+        protein_designation (str): existing protein designation, e.g. "Protein_id" or "Protein"
```
Elena-kal: The doc string does not match the actual parameters - what about  existing_designation and new_designation?

backend/protzilla/importing/cross_linking_import.py
```diff
@@ -0,0 +1,512 @@
...
+            "Protein1": "string",
+            "Protein2": "string",
+            "Is_intra_crosslink": "bool",
+            "Crosslinker": "string",
+            "Peptide1": "string",
+            "Peptide2": "string",
+            "Q_value": "Float64",
+        }
+    )
+    return df.loc[:, columns_in_cross_linking_df]
```
Elena-kal: It is not just here, but let's keep it consistent and always write crosslinking and not cross_linking.

Elena-kal (commented):
Overall, I think this looks good. I added a couple of nitpicky comments. I agree with Joris that the doc strings should be changed and added accordingly. Also tests should be implemented.

backend/protzilla/importing/cross_linking_import.py
NeleRiediger: To be honest, I hadn't considered that option. That was simply my way of making thinks explicit and producing readable code. However, I can be convinced to change it :)

backend/protzilla/importing/cross_linking_import.py
NeleRiediger: I'm aware that this is a bit ugly. But the problem here was, that in the UniProt batch request for some gene names we get newer or more specific versions of this gene names with their respective protein id. Sadly from those responses we have no way of connecting those gene names back to the ones we originally requested. So the gene names from those responses are not in valid_input, as they were never part of the data we initially got.
To catch that I implemented a fallback UniProt request, in which each gene name for which I have no result is queried again individually. Then I get a list of all protein ids connected to those more specific versions of the original gene name. So in this part of the code, the gene_name from the UniProt response would be one of those more specific versions, so it wouldn't be in valid_input, because it was never part of the data we initially got. But if is_fallback is true, I know that I used a fallback request to get this data, which was an individual request for the one gene_name I then stored in valid_input. So I can store the protein ids from this response in the field of the gene_name I used for the request (=valid_input) even though it doesn't match the gene_names of the request's response.

jorisfu: Makes sense i suppose. Perhaps you should document that somewhere in the code explicitly so people don't get confused

NeleRiediger:
> Note: Importing `20230728_SFB_BS3_Gel_MRN_ATM_S3_Xi1.8.7_CSM_xiFDR2.3.5.csv` we were given does not work:
> 
> An error occurred while reading the file: KeyError "Only a column name can be used for the key in a dtype mappings argument. 'Crosslinker' not found in columns.". Please provide a valid cross linking file.

Öhm that's weird, I cannot reproduce this error, it works fine on my machine. I'm not sure what could cause this.

Elena-kal (approved):

PR #197, opened 2026-01-17
196 add persistence option

### LINKED ISSUE #196: Add persistence option for protein prediction load
## Issue description
Instead of saving the prediction data pulled from the Alphafold DB in user data across runs, there should be the option to only load the prediction data for the current run. 

## Acceptance Criteria
- [ ] The user can select whether or not to save the prediction data across runs
- [ ] If the user chooses so, all loaded prediction data is saved across runs
- [ ] If the user chooses so, the loaded predicion data is only saved for this specific run

Elena-kal:
## Description
fixes #196
There now is an option for the user whether to save the data loaded from the alphafold db across runs or only for the current run. Also regardless of the choice, in both cases the data is also loaded into dataframes so we can work with them later down the line.

## Changes
Most important logic is in [alphafold_protein_structure_load.py](backend/protzilla/importing/alphafold_protein_structure_load.py). The option was also added in the forms. I added one library to parse the cif files.

## Testing
Start a new run. Load a specific protein and click the option to save it across runs. It should then be saved under user_data/external_data/alphafold. When uploading the same again, it should not create a new entry or folder. Under the Tables tab you should be able to find all the output dataframes. When not clicking the option to save it across runs, you should still be able to see these dataframes, however the data should not be saved under user_data anymore.

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**

### DIFF
--- a/backend/protzilla/importing/alphafold_protein_structure_load.py
+++ b/backend/protzilla/importing/alphafold_protein_structure_load.py
@@ -6,6 +6,9 @@ from typing import Any
 import pandas as pd
 import requests
 from textwrap import wrap
+from CifFile import ReadCif
+import shutil
+import tempfile
 
 
 from backend.protzilla.constants import paths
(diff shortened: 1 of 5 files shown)


backend/protzilla/importing/alphafold_protein_structure_load.py
```diff
@@ -29,9 +41,244 @@ def _download_file(session: requests.Session, url: str, dest: Path) -> Path | No
...
+    """
+    Convert a protein sequence to FASTA format.
+
+    :param seq: The protein sequence to convert
+    :param header: The header line for the FASTA record (default: "protein_sequence")
+    :param width: The maximum line width for sequence wrapping (default: 60)
+    :return: The sequence in FASTA format
+    :raises ValueError: If the sequence contains invalid characters or whitespace
+    """
+    VALID_AA = set("ACDEFGHIKLMNPQRSTVWYBXZJUO*-")
```
AnnaPolensky: What does AA stand for?

backend/protzilla/importing/alphafold_protein_structure_load.py
```diff
@@ -29,9 +41,244 @@ def _download_file(session: requests.Session, url: str, dest: Path) -> Path | No
...
+    if not seq or any(c.isspace() for c in seq):
+        raise ValueError("Sequence must be a single, whitespace-free string.")
+    seq = seq.upper()
+    bad = set(seq) - VALID_AA
+    if bad:
+        raise ValueError(f"Invalid characters in sequence: {''.join(sorted(bad))}")
+    return ">" + header + "\n" + "\n".join(wrap(seq, width)) + "\n"
```
AnnaPolensky: nit: Maybe we could use a f-string here? Like: f">{header}\n{'\n'.join(wrap(seq, width))}\n"
Both should work, but I think that f-strings are a bit more pythonic?

backend/protzilla/importing/alphafold_protein_structure_load.py
```diff
@@ -29,9 +41,244 @@ def _download_file(session: requests.Session, url: str, dest: Path) -> Path | No
...
+
+        if seq_id is not None:
+            sequence = "".join(seq)
+            records.append(
+                {
+                    "id": seq_id,
+                    "sequence": sequence,
+                    "length": len(sequence),
+                }
+            )
```
AnnaPolensky: nit: this is the same code as in line 83 to 89, maybe extract to a function next time

AnnaPolensky (commented):
Works locally for me as described under Testing. 

Is it consistent how we use internal functions (the ones that start with an underscore) in alphafold_protein_structure_load? We have _download_file and _handle_alphafold_files as the only two internal functions, but for example to_fasta is not internal, although it is only called in _handle_alphafold_files, which is internal

backend/protzilla/importing/alphafold_protein_structure_load.py
```diff
@@ -29,9 +41,244 @@ def _download_file(session: requests.Session, url: str, dest: Path) -> Path | No
         return None
```
jorisfu: I highly recommend putting this function somewhere else, as it's not specific to alphafold downloads. Maybe under a new `backend/protzilla/networking,py`?

jorisfu: Also what's the reason for passing a session instead of creating a new one? Unless we need cookies for AF, I would but all the networking logic possible right into this function

backend/protzilla/importing/alphafold_protein_structure_load.py
jorisfu: probably amino acids?

backend/protzilla/importing/alphafold_protein_structure_load.py
```diff
@@ -29,9 +41,244 @@ def _download_file(session: requests.Session, url: str, dest: Path) -> Path | No
...
+    if not seq or any(c.isspace() for c in seq):
+        raise ValueError("Sequence must be a single, whitespace-free string.")
+    seq = seq.upper()
+    bad = set(seq) - VALID_AA
+    if bad:
+        raise ValueError(f"Invalid characters in sequence: {''.join(sorted(bad))}")
+    return ">" + header + "\n" + "\n".join(wrap(seq, width)) + "\n"
+
+
+def fasta_to_dataframe(fasta_path: str) -> pd.DataFrame:
```
jorisfu: We already have `backend/protzilla/importing/fasta_import.py`, I'd prefer if we reuse that somehow

backend/protzilla/importing/alphafold_protein_structure_load.py
```diff
@@ -29,9 +41,244 @@ def _download_file(session: requests.Session, url: str, dest: Path) -> Path | No
...
+                col_values.append(row[j])
+            else:
+                col_values.append(None)
+        data[col] = col_values
+
+    return pd.DataFrame(data)
+
+
+def _handle_alphafold_files(
+    session: requests.Session,
```
jorisfu: Again I'd prefer not passing sessions around if it's not necessary (put session logic into file download function)

backend/protzilla/importing/alphafold_protein_structure_load.py
```diff
@@ -29,9 +41,244 @@ def _download_file(session: requests.Session, url: str, dest: Path) -> Path | No
...
+    saves metadata to a CSV file.
+
+    :param session: The requests session to use for downloading files
+    :param files_urls: Dictionary containing URLs for CIF, PAE, and pLDDT files
+    :param uniprot: The UniProt ID of the protein
+    :param seq: The protein sequence
+    :param metadata_df: DataFrame containing AlphaFold metadata
+    :param acc: The accession number (used for directory naming)
+    :param persist_upload: If True, files are saved persistently; if False, only loaded into memory
+    :return: Tuple of (cif_df, pae_df, plddt_df, sequence_df) or None values for failed loads
```
jorisfu: Keeping more in line with our current data model I propose this should be a dictionary mapping "cif_df" etc. to the corresponding dataframes

backend/protzilla/importing/alphafold_protein_structure_load.py
Elena-kal: Amino Acid, but I'll rename it :))

backend/protzilla/importing/alphafold_protein_structure_load.py
```diff
@@ -29,9 +41,244 @@ def _download_file(session: requests.Session, url: str, dest: Path) -> Path | No
...
+                    try:
+                        if key == "cifUrl":
+                            try:
+                                cif_df = read_alphafold_mmcif(str(saved))
+                            except Exception:
+                                logger.exception(
+                                    "Failed to load CIF into dataframe. Path=%s",
+                                    str(saved),
+                                )
+                                raise
```
jorisfu: raise what?

backend/protzilla/importing/alphafold_protein_structure_load.py
Elena-kal: Shouldn't be there at the first occurance - makes no sense there. I'll delete it :)) But thanks for the comment

jorisfu (changes requested):

jorisfu:
also please add tests much thx

backend/protzilla/methods/importing.py
```diff
@@ -426,6 +430,11 @@ def create_form(self):
                     name="uniprot",
                     label="Protein ID",
```
jorisfu, line 431: Maybe specify "Uniprot Protein ID"

jorisfu:
Tested working
mynd

backend/protzilla/importing/alphafold_protein_structure_load.py
```diff
@@ -51,58 +26,15 @@ def to_fasta(seq: str, header: str = "protein_sequence", width: int = 60) -> str
     :return: The sequence in FASTA format
     :raises ValueError: If the sequence contains invalid characters or whitespace
     """
-    VALID_AA = set("ACDEFGHIKLMNPQRSTVWYBXZJUO*-")
+    VALID_AMINO_ACID = set("ACDEFGHIKLMNPQRSTVWYBXZJUO*-")
```
jorisfu: plural for set would be more fitting: `VALID_AMINO_ACIDS`

backend/protzilla/importing/alphafold_protein_structure_load.py
jorisfu: Docstring does not match return type

jorisfu (changes requested):

backend/protzilla/importing/alphafold_protein_structure_load.py
Elena-kal: You're right, thanks.

PR #204, opened 2026-01-20
Fix dimensionality reduction

hendraet:
## Description

Fixes dimensionality reduction and corresponding scatter plot so that it is hopefully usable again.

## Changes

- Dimensionality reduction can now work with more than two dimensions
- Updated code for coloring points in scatter plot according to metadata
- Fixed issues that not all of the dim. red. metrics could actually be used
- Error handling was outdated and created cluttered/redundant error messages - fixed
- Scatter plot tests were subpar and were extended
- Extended dimensionality reduction tests as well

## Testing

Could use a run like [this](https://github.com/user-attachments/files/24739896/dim_red_test.zip) to test it (let's see if run import is actually fixed). Of course, parameters should be adapted a bit during testing. 

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**


backend/tests/protzilla/data_analysis/test_dimension_reduction.py
```diff
@@ -87,148 +91,208 @@ def dimension_reduction_four_proteins_df():
 
 
 @pytest.fixture
-def tsne_assertion_df():
```
hendraet, line 90: Explanation: I removed these fixtures (and the tests below) because I couldn't get them to run. They worked on my machine but failed in the pipeline. I guess, setting a seed only helps to some extent when running on different machines.

I decided to at least do some kind of sanity checking and hope that the algorithm itself works fine.

backend/protzilla/data_analysis/plots.py
```diff
@@ -27,67 +27,75 @@
...
 
     :return: returns a dictionary containing a list with a plotly figure and/or a list of messages
     """
+    if isinstance(metadata_df, pd.DataFrame):
+        if metadata_column not in metadata_df.columns:
+            raise ValueError(
+                "The column selected for annotation is not present in the corresponding metadata dataframe.",
+            )
 
-    intensity_df_wide = long_to_wide(input_df) if is_long_format(input_df) else input_df
```
hendraet, line 43: Removed because currently there is no way to even pass a df in long format to the function (when using PROTzilla in a normal way).

backend/protzilla/methods/data_analysis.py
```diff
@@ -1047,36 +1047,53 @@ def create_form(self):
...
+        metadata_field = form["metadata_df"]
 
         input_df_field.set_options(
             form_helper.to_choices(
                 run.steps.get_instance_identifiers(
                     DimensionReductionUMAP, "embedded_data"
                 )
+                + run.steps.get_instance_identifiers(
+                    DimensionReductionTSNE, "embedded_data"
+                )
```
tE3m: this would catch all steps that have `embedded_data` as an output key, maybe that's better than specifying? In the refactor I'm removing the instance checks anyways
```suggestion
                run.steps.get_instance_identifiers(
                    Step, "embedded_data"
                )
```

backend/protzilla/methods/data_analysis.py
```diff
@@ -1047,36 +1047,53 @@ def create_form(self):
...
 
     # TODO: input
     def insert_dataframes(self, steps: StepManager, inputs) -> dict:
         inputs["input_df"] = steps.get_step_output(
             Step, "embedded_data", inputs["input_df"]
         )
-        inputs["color_df"] = steps.get_step_output(Step, "color_df", inputs["color_df"])
+        inputs["metadata_df"] = steps.get_step_output(
+            Step, "metadata_df", inputs["metadata_df"]
+        )
```
tE3m: There's a property defined for `metadata_df` and `protein_df`
```suggestion
        inputs["metadata_df"] = steps.metadata_df
```

backend/tests/protzilla/data_analysis/test_dimension_reduction.py
tE3m: That's odd, but I guess not worth debugging

backend/protzilla/data_analysis/dimension_reduction.py
tE3m: From what I can see, this method is never set and defaults to `barnes_hut`. While that has been the behavior before this PR as well, maybe now would be the time to add the dropdown in the form, or is this intended?

tE3m (commented):
Some questions, but the core changes look good and sensible

tE3m:
Also, a general question regarding the `Plot` steps: is this still the right approach? It looks to me like plotting in separate steps was the "old" way and all modernized steps have the plotting integrated via the `plot_method` attribute. Of course that's way out of scope for this PR, but as a general question: Should the plotting steps be merged into their relevant calculation steps?

backend/protzilla/methods/data_analysis.py
hendraet: Somehow, I thought it would be better to be specific here. I guess the reasoning was that it was surely tested with these steps, and if different steps coincidentally also produce embedded data (which is somehow not meant for visualization), we won't run into errors. But this is probably overly cautious.

backend/protzilla/data_analysis/dimension_reduction.py
hendraet: I didn't see a reason not to give the user the choice as it is unlikely that they deal with millions of data points, so I quickly implemented it. Please have a look.

hendraet:
> Also, a general question regarding the `Plot` steps: is this still the right approach? It looks to me like plotting in separate steps was the "old" way and all modernized steps have the plotting integrated via the `plot_method` attribute. Of course that's way out of scope for this PR, but as a general question: Should the plotting steps be merged into their relevant calculation steps?

You are correct, there are basically no steps that do not provide their own plotting method (volcano plot being one of the few examples). 
In general, I would agree that specific plots should be directly tied to their resepective step, but the scatter plot should work with multiple different steps and not just the dim. reduction. I say "should" because this is currently not the case. However, it would be desirable to use the output of clustering steps to color scatter plots. (I'm pretty this was a working feature in the past, but was not maintained properly)

backend/protzilla/data_analysis/dimension_reduction.py
```diff
@@ -72,19 +72,23 @@ def t_sne(
             f"{min(intensity_df_wide.shape[0], intensity_df_wide.shape[1])} (the smaller one of number of "
             "samples/features). "
         )
-    if n_components > 3 and method == "barnes_hut":
+    if n_components > 3 and method == "Barnes-Hut approximation":
```
tE3m: I know that this is nitpicky, but maybe the strings should also come from the enum?

tE3m (commented):
Overall very good, just one tiny request

backend/protzilla/data_analysis/dimension_reduction.py
hendraet: You are absolutely right. Fixed it

tE3m (approved):
not sure if not implicitly clear, but I'm happy with the changes as they are

PR #208, opened 2026-01-22
Pull example data from pride

### LINKED ISSUE #165: Find a different way to store the example dataset
## Issue description

It's currently not possible to download the Example Dataset (maybe also because of the CI?) because the repository exceeds its git lfs budget. From the error log:
```
Error downloading object: backend/user_data/example_dataset/evidence.txt (ba2e8cc): Smudge error: Error downloading backend/user_data/example_dataset/evidence.txt (ba2e8cc7cd02ba78deb30228e2b0158afb90e9047d0f904b7002401de0dd9e6b): batch response: This repository exceeded its LFS budget. The account responsible for the budget should increase it to restore access.
```
Maybe we can host it somewhere else and download it during installation. 

## Acceptance Criteria
- [ ] Find a workaround so that the example dataset is also easily accessible for users, i.e., does not need any extra clicks.

hendraet:
## Description

fixes #165 (kinda)
Removes the data for the example dataset from git lfs. Data will now be downloaded directly from PRIDE during the first import of the example dataset.

## Changes

Mainly changed `example_dataset_import.py`. and introduced the import via pridepy. I also removed the example dataset test completely because I didn't see a way to test the data loading without downloading a 2.5GB archive every time (which would probably put unnecessary strain on our CI pipeline).
I also fixed some issues regarding the peptide import that I encountered along the way. 

## Testing

Just execute the first step (and maybe also subsequent steps) of the example workflow. First time might take a while because the dataset will be downloaded.
Also, execute the test locally because the integration test in `test_runner.py` will not run in the CI but only locally.

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**


backend/protzilla/form_helper.py
```diff
@@ -4,7 +4,7 @@
 
 
 def to_choices(choices: list[str], required: bool = True) -> list[Option]:
-    return (
+    return sorted(
```
hendraet, line 7: I mainly added this because a list of proteins in one of the steps was all over the place, and it was quite hard to find what I was looking for. Not really necessary for this PR, but I thought that I'd leave it in.

backend/protzilla/importing/peptide_import.py
```diff
@@ -122,56 +136,108 @@ def peptide_import(file_path: Path, intensity_name: str, map_to_uniprot) -> dict
...
+            "Leading razor protein",
+            "Sequence",
+            intensity_name,
+            "Modifications",
+            "Modified sequence",
+            "Missed cleavages",
+            "Experiment",
+            "PEP",
+            "Raw file",
+        ]
```
tE3m: These should probably be defined somewhere in `protzilla/constants` unless there's a good reason not to

backend/protzilla/importing/peptide_import.py
```diff
@@ -122,56 +136,108 @@ def peptide_import(file_path: Path, intensity_name: str, map_to_uniprot) -> dict
...
+                        msg=f"{intensity_name} was not found in the provided file, please use another intensity measure "
+                        "and try again or verify your file.",
+                    )
+                ],
+            )
+
+        df = df.rename(
+            columns={
+                c: c.capitalize() if c not in column_exceptions and " " in c else c
+                for c in df.columns
```
tE3m, line 190: feels like I've seen something like this before - might be good to introduce a general method that does this (but not needed for this PR)

.github/workflows/backend_ci.yml
```diff
@@ -54,12 +54,12 @@ jobs:
       - name: Run tests and build coverage file
         run: |
           export PYTHONPATH=. pytest 
-          pytest --cache-clear --cov=. --cov-report term-missing backend/tests/
+          pytest --cache-clear --cov=. --cov-report term-missing backend/tests/ --deselect=backend/tests/protzilla/test_runner.py::test_example_dataset_runner
```
tE3m: I feel like it would be cleaner to mark a test to be skipped in the test file instead of here - like it's described [here](https://dev.to/sql_knievel/how-to-skip-a-pytest-test-when-running-in-github-actions-1m2o)

tE3m (commented):
Code looks good, I wasn't able to actually perform the download since I'm on a slow network right now, but I will return to once the connection is better

backend/protzilla/importing/example_dataset_import.py
```diff
@@ -1,23 +1,83 @@
...
+                protocol="ftp",
+                username=None,
+                password=None,
+                aspera_maximum_bandwidth=None,
+                checksum_check=False,
+            )
+        except Exception as e:
+            raise RuntimeError(
+                f"Error downloading file {filename} from PRIDE project {accession}. "
+                f"This is likely and issue with PRIDE.\nOriginal error: {e}\n"
```
Elena-kal:
```suggestion
                f"This is likely an issue with PRIDE.\nOriginal error: {e}\n"
```

tE3m:
I also just saw that there is apparently progress bar (was only displayed once I shut down the server) - maybe, this could be used to display some sort of progress to the user? Not critical, but certainly nice to have

backend/protzilla/importing/example_dataset_import.py
```diff
@@ -1,23 +1,83 @@
...
-def example_dataset_import():
-    intensity_name = IntensityType.LFQ_INTENSITY.value
+def download_example_data(
+    import_peptide_data: bool,
+    accession: str = "PXD014997",
+    filename: str = "SEARCH-IDENTIFICATIONS_REL_FREE-RELAPSE.7z",
+) -> None:
+    archive_file_path = EXAMPLE_DATASET_DIR / filename
+    if not archive_file_path.exists():
+        logger.info(f"Downloading file %s from PRIDE project %s", filename, accession)
```
Elena-kal:
```suggestion
        logger.info(f"Downloading file {filename} from PRIDE project {accession}")
```

backend/protzilla/importing/peptide_import.py
```diff
@@ -122,56 +136,108 @@ def peptide_import(file_path: Path, intensity_name: str, map_to_uniprot) -> dict
...
+            IntensityType.RATIO_LH_NORMALIZED.value,
+        }
+
+        def select_column(column):
+            capitalized_column = (
+                column.capitalize()
+                if column not in column_exceptions and " " in column
+                else column
+            )
+            return capitalized_column in id_columns
```
Elena-kal, line 166: could we rename this function? I feel like the name is a bit misleading as we are not just selecting columns but also capitalizing them.

backend/protzilla/importing/example_dataset_import.py
Elena-kal:
```suggestion
        logger.info(f"Downloading file {filename} from PRIDE project {accession}")
```

backend/protzilla/importing/example_dataset_import.py
Elena-kal:
```suggestion
                f"This is likely an issue with PRIDE.\nOriginal error: {e}\n"
```

backend/protzilla/methods/importing.py
```diff
@@ -395,15 +401,35 @@ class ExampleDatasetImport(ImportingStep):
...
+    def modify_form(self, form, run):
+        import_peptide_data_field = form["import_peptide_data"]
+        if import_peptide_data_field.value is False:
+            run.steps.current_step.output_keys = ["metadata_df", "protein_df"]
+        else:
+            run.steps.current_step.output_keys = [
+                "metadata_df",
+                "peptide_df",
+                "protein_df",
+            ]
```
Elena-kal, line 435:
```suggestion
        if import_peptide_data_field.value is True:
            run.steps.current_step.output_keys = [
                "metadata_df",
                "peptide_df",
                "protein_df",
            ]
```

backend/tests/protzilla/importing/test_example_data_import.py
```diff
@@ -0,0 +1,173 @@
...
+def tmp_example_data_dir(tmp_path_factory):
+    test_tmp_data_dir = Path("example_data/")
+    tmp_path = tmp_path_factory.mktemp(str(test_tmp_data_dir))
+    return tmp_path
+
+
+@pytest.fixture()
+def example_data_paths(tmp_example_data_dir):
+    tmp_protein_path = (
+        tmp_example_data_dir.resolve() / "txt_REL_FREE-REPASE/proteinGroups.txt"
```
Elena-kal: RELAPSE instead of REPASE? if not, what is REPASE?

backend/tests/protzilla/importing/test_example_data_import.py
```diff
@@ -0,0 +1,173 @@
...
+    )
+    tmp_evidence_path = (
+        tmp_example_data_dir.resolve() / "txt_REL_FREE-REPASE/evidence.txt"
```
Elena-kal: here as well

backend/tests/protzilla/test_runner.py
```diff
@@ -401,6 +406,57 @@ def test_integration_runner(
     assert_runner_finished_successfully(runner)
 
 
+def test_example_dataset_runner(tests_folder_name, monkeypatch):
+    protein_file = EXAMPLE_DATASET_DIR / "txt_REL_FREE-REPASE/proteinGroups.txt"
```
Elena-kal: REPASE?

Elena-kal (commented):
I remarked some minor changes and typos. Overall the code looks good.
I could not test the workflow because the data would take too long to download which is definetly my problem because I have terrible WIFI and only weak VPN that can take the data through the Chinese Firewall. I hope Tarek tested the step well. (Sorry :)))

Elena-kal (commented):
I remarked some minor changes and typos. Overall the code looks good.
I could not test the workflow because the data would take too long to download which is definetly my problem because I have terrible WIFI and only weak VPN that can take the data through the Chinese Firewall. I hope Tarek tested the step well. (Sorry :)))

.github/workflows/backend_ci.yml
hendraet: The idea was to skip the test only in the runners to not trigger the download every single time. However, if you have the files on disk anyway, it would be nice to run this test locally. This way, you can ensure that the example workflow actually runs, as this runner is the closest to an integration test that we have.

I was afraid that just skipping it, would lead to it never being executed anymore.

hendraet:
> I also just saw that there is apparently progress bar (was only displayed once I shut down the server) - maybe, this could be used to display some sort of progress to the user? Not critical, but certainly nice to have

You mean as an actual UI element? If yes, then I would not do it in this PR (to wrap this one up and also because I'm currently a bit swamped with work). But, this might be a good issue for the hackathon :wink:

backend/protzilla/importing/example_dataset_import.py
hendraet: This is intentional. I made it a habit because pylint advises against usage of f-strings in logging statements: https://pylint.readthedocs.io/en/stable/user_guide/messages/warning/logging-fstring-interpolation.html

backend/protzilla/methods/importing.py
hendraet: But wouldn't that make it impossible to revert the toggling, as soon as the checkbox is unchecked again?

Elena-kal: Oh true! Did not think of this... Then it's perfect just the way it is :))

backend/tests/protzilla/importing/test_example_data_import.py
hendraet: Yeah, I was just trying to keep it consistent with the original dataset. This is literally what the subfolder is called in the zip archive that is pulled from PRIDE.

What one could do though, is not hardcode this string here, but construct it based on `EXAMPLE_DATASET_PROTEIN_FILE` in [backend/protzilla/constants/paths.py](https://github.com/cschlaffner/PROTzilla/pull/208/changes/BASE..7643a8c21bfd18fb5fcf577dfb552f3184d0a6e8#diff-e65bc973343cadaf624c4c35ac14f4ac9f91d4b5ba571348417bfbfb7c510542)

backend/protzilla/importing/example_dataset_import.py
Elena-kal: but it is still an f string right? but i guess then it should be fine 👍

.github/workflows/backend_ci.yml
tE3m: Yes, I got that - I feel like instead of hardcoding this in the CI workflow, one could check if the env var `GITHUB_ACTIONS` is set and then use `@pytest.mark.skipif(env_var)` to skip the test depending on this condition. The effect is the same, but it feels more idiomatic to me to do this in the test file and not in the workflow file

tE3m:
> > I also just saw that there is apparently progress bar (was only displayed once I shut down the server) - maybe, this could be used to display some sort of progress to the user? Not critical, but certainly nice to have
> 
> You mean as an actual UI element? If yes, then I would not do it in this PR (to wrap this one up and also because I'm currently a bit swamped with work). But, this might be a good issue for the hackathon 😉

A UI element would of course be optimal, but overkill for now. I just thought of e.g. logging the progress through messages. But right now I'm not even sure if it's possible to output messages outside of the return values from calc methods, so maybe that wouldn't be so easy.
I think that ultimately, there should be some progress feedback to the users (2.5G of blocking download is not exactly perfect UX), but of course, non-blocking would require THRÄÄDS (@jorisfu) - which is somewhere down the road on my mental roadmap

.github/workflows/backend_ci.yml
hendraet: Yes, sounds much cleaner, will do. I just didn't know that this was an option

backend/protzilla/importing/example_dataset_import.py
hendraet: Oh yeah, I didn't see the leftover "f". Will remove

backend/protzilla/importing/peptide_import.py
hendraet: The function does not actually capitalize the column name in the dataframe. It's just temporary for the selection. The real capitalization happens only as the dataframe was read and columns selected

backend/tests/protzilla/importing/test_example_data_import.py
hendraet: I kept the original string in the `paths.py` file to avoid extra logic to rename the extracted directory, but I created all other paths dynamically to avoid multiple, explicit Typos in the code

hendraet:
> > > I also just saw that there is apparently progress bar (was only displayed once I shut down the server) - maybe, this could be used to display some sort of progress to the user? Not critical, but certainly nice to have
> > 
> > 
> > You mean as an actual UI element? If yes, then I would not do it in this PR (to wrap this one up and also because I'm currently a bit swamped with work). But, this might be a good issue for the hackathon 😉
> 
> A UI element would of course be optimal, but overkill for now. I just thought of e.g. logging the progress through messages. But right now I'm not even sure if it's possible to output messages outside of the return values from calc methods, so maybe that wouldn't be so easy. I think that ultimately, there should be some progress feedback to the users (2.5G of blocking download is not exactly perfect UX), but of course, non-blocking would require THRÄÄDS (@jorisfu) - which is somewhere down the road on my mental roadmap

I totally get. The UX is "suboptimal". I noted it down as a potential issue/improvement that could be tackled during a hackathon.

tE3m:
The previously `py7zr` version specified in requirements was [yanked](https://pypi.org/project/py7zr/1.1.2/), so I changed the version specifier. I've also been able to functionally test the PR - downloading took like half an hour despite a pretty solid connection, so idk if that's down to pride's file servers.

tE3m (approved):
lgtm

hendraet:
> The previously `py7zr` version specified in requirements was [yanked](https://pypi.org/project/py7zr/1.1.2/), so I changed the version specifier. I've also been able to functionally test the PR - downloading took like half an hour despite a pretty solid connection, so idk if that's down to pride's file servers.

Thanks for the fix.

The download speed doesn't sound too good, but not bad either. For the 3GB file that was around 2MB/s.
Unfortunately, we have to download the full zip file even though we only need some of the files.

PR #212, opened 2026-01-29
191 validate cross links based on armstrong error tolerance

### LINKED ISSUE #191: Validate Cross-Links based on Armstrong-Error-Tolerance
## Issue description
The user should be able to enter by how many Armstrong the Cross-Link length is allowed to differ from the by AlphaFold predicted distance between the two connected amino acids.

## Acceptance Criteria
- [ ] User can enter the cross-linker length for each cross-linker.
- [ ] User can enter the maximum difference they allow between predicted distance and actual cross-linker length in Armstrong.
- [ ] Output of a simple diagram showing how many cross-linkers are within the error-tolerance and how many are not.
- [ ] Integrated a suitable library for determining the distances between two amino acids in the AlphaFold .civ/calculating distances based on the central C-atom.

AnnaPolensky:
## Description
fixes #191 
There is a new step where the intra-crosslinks of one protein are being validated (-> the user can enter an allowed deviation from the crosslinker's length. If the distance of the connected amino acids in alphafold is greater than crosslinker length + allowed deviation, the crosslinker will be marked as invalid. The user can not only apply an upper bound but also a lower bound to the allowed deviation.)
Right now, the distance calculation is based on the central Ca atom; we might want to change this to the residue atom where the crosslinker actually binds later on.
The output is a simple bar plot and we add two colums to the crosslinking df (distance in alphafold and true or false on whether the crosslink matches the alphafold data or not).
grafik

## Changes
All the functions for the actual validation are in backend/protzilla/data_analysis/cross_linking_validation.py
The form is defined at the end of backend/protzilla/methods/data_analysis.py
Simple tests are in backend/tests/protzilla/data_analysis/test_crosslinking_validation.py

## Testing

1. Add Crosslinking Data Import and import p26splussubstrate_XL.xlsx
2. Add Alphafold Load Step and load O43242
3. Add Crosslinking Validation (this is a data analysis step), select O43242, DSSO should have a length of 10.2. Test different values for the bounds. E.g. if we only apply an upper bound of 1, only one crosslink should match the Alphafold data, for 2, 2 crosslinks should match.

Check that in the crosslinking validation step three fields show up for each crosslinker that appears in the crosslinking data (for p26splussubstrate_XL.xlsx only DSSO should show up).

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**


backend/protzilla/methods/data_analysis.py
```diff
@@ -2457,3 +2462,77 @@ def create_form(self):
...
+                    name="protein_to_validate",
+                    label="Protein prediction that should be validated",
+                ),
+            ],
+        )
+
+    def modify_form(self, form: Form, run: Run) -> None:
+        cross_linker = self._get_crosslinker_names_from_crosslinker_df(run.steps)
+        for cl in cross_linker:
+            field_name = f"length_of_{cl}"
```
Elena-kal:
```suggestion
            field_name = f"{cl}_length"
```
super nitpicky though, just feels better but I don't know what would be considered 'correct'

backend/protzilla/methods/data_analysis.py
```diff
@@ -2457,3 +2462,77 @@ def create_form(self):
...
+            if field_name not in form:
+                crosslinker_length_field = FloatField(
+                    name=field_name,
+                    label=f"Length of {cl} in Ångström",
+                    min=0,
+                )
+                upper_bound_length_deviation_field = FloatField(
+                    name=f"upper_accepted_deviation_for_{cl}",
```
Elena-kal:
```suggestion
                    name=f"{cl}_upper_accepted_deviation",
```
same here, not sure but it feels more aligned with how the names are used in the past forms

backend/protzilla/methods/data_analysis.py
```diff
@@ -2457,3 +2462,77 @@ def create_form(self):
...
             input_fields=_PTMVisualizationWithGroups.get_form_fields(),
         )
+
+
+class CrossLinkingValidationWithAngstromDeviation(DataAnalysisStep):
+    display_name = "Ångström Deviation"
+    operation = "Cross Linking Validation"
+    method_description = "Validates cross links based on the difference between the length of the cross linker and the distance between the amino acids which were connected by the cross linker. (in Ångström)"
+
+    output_keys = ["crosslinking_df_result"]
```
Elena-kal:
```suggestion
    output_keys = ["crosslinking_result_df"]
```

backend/protzilla/data_analysis/crosslinking_validation.py
Elena-kal, line 1: I am a bit confused about the naming of the file. Did we agree on whether it should be cross-linking or crosslinking? Because right now I do not think we keep it consistent across the code base...

backend/protzilla/data_analysis/crosslinking_validation.py
```diff
@@ -0,0 +1,222 @@
...
+    fetch_alphafold_protein_structure,
+)
+from protzilla.data_preprocessing.plots import create_bar_plot
+
+
+def get_reactive_atom_of_amino_acid_residue(amino_acid_kind: str) -> str:
+    # right now we always return the central C atom
+    # later we might want to return the reactive atom of the amino acid residue of the specific amino acid kind
+    # as soon as we change this, we will need to change the test test_validate_with_angstrom_deviation
+    return "CA"
```
Elena-kal, line 23: I was just wondering: How do we distinguish between the same kinds of atoms at different positions? Say we have another C atom, how do we know which one is the central?

backend/protzilla/data_analysis/crosslinking_validation.py
```diff
@@ -0,0 +1,222 @@
...
+        & (
+            cif_df["_atom_site.label_seq_id"].astype(int)
+            == amino_acid_position_where_crosslinker_bound
+        )
+    ]
+
+    if cif_df.empty:
+        raise ValueError(
+            f"No {relevant_atom} atom found for amino acid at position {amino_acid_position_where_crosslinker_bound}."
+        )
```
Elena-kal, line 57: Maybe raise a more specific error here because we do not even have a cif file where we could find an atom in.

backend/protzilla/data_analysis/crosslinking_validation.py
Elena-kal, line 1: I think half of the functions are missing doc strings.

backend/protzilla/data_analysis/crosslinking_validation.py
```diff
@@ -0,0 +1,222 @@
...
+
+    return distance
+
+
+def get_position_of_amino_acid_crosslinker_bound_to(
+    protein_sequence: str,
+    peptide_sequence: str,
+    crosslinker_position_within_peptide: int,
+) -> int:
+    """Returns which amino acid the cross-linker bound to, 1-based."""
```
Elena-kal: Maybe make the doc string more consistent with the rest of the doc strings in our code base

backend/protzilla/data_analysis/cross_linking_validation.py
```diff
@@ -0,0 +1,222 @@
...
+    peptide_sequence: str,
+    crosslinker_position_within_peptide: int,
+) -> int:
+    """Returns which amino acid the cross-linker bound to, 1-based."""
+    peptide_start_position = protein_sequence.find(peptide_sequence) + 1
+    if peptide_start_position == 0:
+        raise ValueError(
+            f"Peptide {peptide_sequence} was not found in protein sequence"
+        )
+    return peptide_start_position + crosslinker_position_within_peptide - 1
```
Elena-kal: I do not understand the reason for first adding 1 checking for `peptide_start_position == 0` and then later subtracting 1 again. Why don't we simply not add 1, check for  `peptide_start_position == -1` and don't subtract in the end. Feels much more intuitive and is less adding and subtracting.

backend/protzilla/data_analysis/cross_linking_validation.py
```diff
@@ -0,0 +1,222 @@
...
+
+
+def get_distance_between_crosslinker_connected_amino_acids_in_alphafold(
+    fasta_df: pd.DataFrame, cif_df: pd.DataFrame, crosslink
```
Elena-kal: what is the crosslink? Is it the name? maybe add the type or make the variable name more descriptive.

backend/protzilla/data_analysis/cross_linking_validation.py
```diff
@@ -0,0 +1,222 @@
...
+    AlphaFold data or not
+    :raises KeyError: If a required crosslinker field is missing in crosslinker_information.
+    :raises ValueError: If peptide sequences cannot be matched to the protein sequence.
+    """
+    alphafold_data = fetch_alphafold_protein_structure(
+        uniprot_id=protein_to_validate, persist_uploads=False
+    )
+    cif_df = alphafold_data["cif_df"]
+    fasta_df = alphafold_data["sequence_df"]
+    df = crosslinking_df.copy()
```
Elena-kal: This is fine for this issue - but we should not pull the protein from alphafold in here as there is an extra step for it and one can upload own structure predictions in the settings soon and might want to use them. Instead I would argue we should add an extra field in the validation form where one can chose which protein structure one wants validate between all the ones that are already uploaded or were added only for this specific run.

Elena-kal: Also maybe make the variable name df a bit more descriptive :))

backend/protzilla/data_analysis/cross_linking_validation.py
```diff
@@ -0,0 +1,222 @@
...
+            accepted_distance_lower_bound <= distance <= accepted_distance_upper_bound
+        )
+
+        return pd.Series({"alphafold_distance": distance, "valid_crosslink": valid})
+
+    df.loc[mask, ["alphafold_distance", "valid_crosslink"]] = (
+        relevant_crosslinks_df.apply(check_crosslink, axis=1)
+    )
+
+    df = df[df["valid_crosslink"].notna()]
```
Elena-kal: naming please 🥲

backend/protzilla/data_analysis/cross_linking_validation.py
```diff
@@ -0,0 +1,222 @@
...
+        try:
+            (
+                crosslinker_length,
+                accepted_deviation_upper_bound,
+                accepted_deviation_lower_bound,
+            ) = crosslinker_information[crosslink.Crosslinker]
+        except KeyError as e:
+            missing_key = e.args[0]
+            raise KeyError(
+                f"Missing required field '{missing_key}' for crosslinker '{crosslink.Crosslinker}'."
```
Elena-kal: Seems weird to call it missing field when its a key thats missing right? So either the variable naming is off or maybe rewrite the error message.

Elena-kal (commented):
Seems fine overall. The test as you described worked. Although I think because you fetch the alphafold data in your code yourself one does not need to add the step to upload it. I think in future this should not be included in the code but I am fine with leaving it as it is for now, if you open an issue for it as soon as this is merged.

backend/protzilla/data_analysis/crosslinking_validation.py
AnnaPolensky: For each amino acid there is exactly one atom called "CA" (for C alpha). This is the central C atom. The other C atoms are named differently.

backend/protzilla/data_analysis/crosslinking_validation.py
AnnaPolensky: What information are you missing exactly? Because I assumed that if there is no cif file existing for the protein, the alphafold-fetch-function would raise an error.

backend/protzilla/data_analysis/crosslinking_validation.py
AnnaPolensky: When we have some description that shows up for the user, we agreed to use cross-linking, for variable names and file names we want to use crosslinking. Changed the filename

backend/protzilla/data_analysis/cross_linking_validation.py
AnnaPolensky: I thought it would be more consistent to have the peptide_start_position 1-based since the crosslinker_position_within_peptide is 1-based, but then if we have the protein peptide1=A, peptide2=A, peptide3=ABCK and the crosslinker bound to K -> crosslinker_position_within_peptide = 4, we would end up at 3+4=7 although it's the 6th amino acid the crosslinker bound to. But changed it like you suggested

backend/protzilla/data_analysis/cross_linking_validation.py
AnnaPolensky: Does it make more sense with the type or do you have a naming suggestion which would make more sense to you?

backend/protzilla/data_analysis/cross_linking_validation.py
AnnaPolensky: I will open an issue for this

backend/protzilla/data_analysis/crosslinking_validation.py
AnnaPolensky: All of them have docstrings now, but I would discuss with the team whether we really want this, since several of the functions do now have docstrings as long as the function's code or even longer, so I'm unsure whether it is worthwhile to write comments for short functions.

backend/protzilla/data_analysis/crosslinking_validation.py
Elena-kal: Okay I see, what about other atoms like Hydrogen? There are probably several of them in one amino acid right? Do they also have definite naming? No need to change anything here but I 'd be interested to know. Maybe we can talk about it on Tuesday.

backend/protzilla/data_analysis/cross_linking_validation.py
Elena-kal: I think to know the type should be enough

Elena-kal (approved):

backend/protzilla/data_analysis/crosslinking_validation.py
```diff
@@ -0,0 +1,275 @@
...
+    :return: the distance between the two residues in Ångström
+    """
+    x1, y1, z1 = get_coordinates_of_atom_crosslinker_bound_to(
+        amino_acid_position1, amino_acid_kind1, cif_df
+    )
+    x2, y2, z2 = get_coordinates_of_atom_crosslinker_bound_to(
+        amino_acid_position2, amino_acid_kind2, cif_df
+    )
+
+    distance = math.sqrt((x2 - x1) ** 2 + (y2 - y1) ** 2 + (z2 - z1) ** 2)
```
jorisfu: I'd personally prefer if we have the positions be vectors as `np.array` types and calculate the distance via `np.linalg.norm`. Pretty sure this works as well, but numpy is cleaner and probably more accurate: https://www.geeksforgeeks.org/python/calculate-the-euclidean-distance-using-numpy/

backend/protzilla/data_analysis/crosslinking_validation.py
```diff
@@ -0,0 +1,275 @@
...
+            raise KeyError(
+                f"Missing required information regarding crosslinker length "
+                f"and/or accepted deviation for crosslinker '{crosslink.Crosslinker}'."
+            )
+        # Fallback to default deviation bounds when not explicitly provided
+        accepted_distance_lower_bound = crosslinker_length - (
+            accepted_deviation_lower_bound or crosslinker_length
+        )
+        accepted_distance_upper_bound = (
+            accepted_deviation_upper_bound or 1e9
```
jorisfu: Can we make sure `1e9` is practically infinity?

jorisfu (changes requested):
Just a little thing but looks very good

backend/protzilla/data_analysis/crosslinking_validation.py
AnnaPolensky: I'm unsure, I would talk to Chris about this on Tuesday.

PR #214, opened 2026-01-29
Refactor crosslinking import

### LINKED ISSUE #194: Display the failed rows after a cross linking import

### LINKED ISSUE #213: Add organism id input field to cross linking input
## Issue description
For processing the cross linking data correctly we need to know the organism the data belongs to. Therefore a user should be able to indicate the organism id along with uploading the file with the cross linking data. This input should then be validated against the NCBI Taxonomy ids, before being used for the correct uniprot lookup. 

## Acceptance Criteria
- [x] there is an input textfield for the organism id in the crosslinking input step 
- [x] the data from this input is then validated with data from the NCBI Taxonomy ids
- [x] if the indicated organism id could not be validated, the process is terminated with an adequate error message
- [x] if the indicated organism id was successfully validated, it is used in the respective uniprot request, to process the cross linking data properly

NeleRiediger:
## Description
fixes #194 
fixes #213 
Apart from those two features some serious refactoring of the import happened: 
- better use of the uniprot api made fallback lookups when getting protein ids from gene names unnecessary
- capping batch requests to a maximum size of 25 inputs, made fallback lookups when getting gene names from protein ids unnecessary
- Isoforms are now kept in the protein id column whenever they are part of the input data, but not requested from uniprot 
- both uniprot requests now return a response in tsv format, which made it possible to combine larger parts of both requests into a single unified function 

## Changes
backend/protzilla/methods/importing.py in order to add the organism id textfield (and output the failed_df)
backend/protzilla/importing/crosslinking_import.py everything else 

## Testing
- test wether you can import a .csv and a .xslx crosslinking file 
- look if the rows/ proteins that could not be processed seem plausible 
- test what happens if you input different organism ids in the respective text field 

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**


backend/protzilla/importing/crosslinking_import.py
AnnaPolensky, line 78: I believe that we don't need the types here since we already have the type hints in the function header.

backend/protzilla/importing/crosslinking_import.py
```diff
@@ -71,15 +76,34 @@ def validate_data_before_lookup(
...
+    The final batch may contain fewer elements.
+
+    :param data: Iterable containing input elements to be batched
+    :type data: iterable
+
+    :return: Iterator yielding batches of input elements as lists
+    :rtype: iterable
+
+    :yields: Lists of at most 25 elements
+    :yield type: list
```
AnnaPolensky, line 110: Same as above

(I don't want to comment on this for all of the functions, so please just check the docstrings of all the functions in this file, in case we don't need the types in the docstrings as well.)

backend/protzilla/importing/crosslinking_import.py
```diff
@@ -167,127 +192,121 @@ def execute_uniprot_request(url, params, valid_data, results):
...
+                    if gene_name in input_data:
+                        results[gene_name] = (True, requested_data, None)
+                        break
+
+
+def uniprot_lookup(
+    input_data: set[str],
+    mode: Literal["id_to_gene_name", "gene_name_to_id"],
+    results: dict[str, tuple[bool, Optional[str], Optional[str]]],
+    organism_id: Optional[str] = None,
```
AnnaPolensky, line 266: I know you are calling this function at some point with organism_id = None, but I'm a bit confused: every time I don't enter an organism ID in protzilla, I get an error message. So why are we still doing uniprot lookups without an organism id?

backend/protzilla/importing/crosslinking_import.py
```diff
@@ -302,132 +321,83 @@ def get_gene_name_from_protein_ids(protein_ids: set):
...
-        r")$"
+        r")"
+        r"(?:-.+)?$"
     )
 
     valid_ids, results = validate_data_before_lookup(
-        protein_ids,
+        data_for_lookup=protein_ids,
         validator_function=lambda pid: bool(valid_id_pattern.match(pid)),
         error_code="NOT_A_VALID_PROTEIN_ID",
```
AnnaPolensky: I think there were some error codes earlier like NO_PROTEIN_ID_FOUND. If these error codes are in the same category, does it make sense to extract them into an enum?

backend/tests/protzilla/importing/test_crosslinking_import.py
```diff
@@ -22,9 +21,10 @@ def test_aggregate_data():
     assert result == {"A", "B", "C", "D"}
 
 
-def test_remove_isoform_from_protein_id():
-    assert remove_isoform_from_protein_id("P12345-2") == "P12345"
-    assert remove_isoform_from_protein_id("Q67890") == "Q67890"
+def test_isoform_removal_logic():
+    ids = {"P12345-2", "Q67890"}
+    cleaned = {x.split("-", 1)[0] for x in ids}
+    assert cleaned == {"P12345", "Q67890"}
```
AnnaPolensky: This test will never fail due to changes in our code. Therefore, it does not really make sense to me, and I would remove/replace it.

backend/tests/protzilla/importing/test_crosslinking_import.py
```diff
@@ -93,18 +93,65 @@ def test_execute_uniprot_request_timeout():
...
+    )
+
+    results = {}
+    uniprot_lookup(
+        input_data={"P1"},
+        mode="id_to_gene_name",
+        results=results,
+    )
+
+    assert results["P1"] == (False, None, "NO_GENE_NAME_FOUND")
```
AnnaPolensky: At first, I wanted to suggest, that these two tests share a lot of code. and that I'm unsure, but that you might want to look into whether you can parameterise the test using @pytest.mark.parametrize?
However, after comparing them more thoroughly I believe that they are testing exactly the same thing?

AnnaPolensky (commented):
There are organism IDs for which I can import the p26splussubstrate_XL.xlsx file, like 1 or 1234, but importing the csm file fails for these organism ids.

Image

backend/protzilla/importing/crosslinking_import.py
NeleRiediger: The uniprot lookups without an organism id are necessary for the .xlsx files, as they don't need an organism id (those files include protein ids, which are unique over all organisms). However I changed that the organism id is a necessary input for all uploads. So now it should be possible to import .xlsx files without entering an organism id.

NeleRiediger:
> There are organism IDs for which I can import the p26splussubstrate_XL.xlsx file, like 1 or 1234, but importing the csm file fails for these organism ids.
> 
> Image

That is expected behavior, because the p26splussubstrate_XL.xlsx file doesn't need any organism ids, while the csm does. These ids are valid ids but probably they don't correspond to any data in the given csm file. That's why the import fails entirely. 
I added more explicit error handling and messages to hopefully make this more user friendly.

backend/protzilla/importing/crosslinking_import.py
NeleRiediger: I didn't even think of that option. Refactored it and introduced enums for most of the strings I use :)

AnnaPolensky: Nice :)

3dot141592 (approved):
Overall, everything looks good. I tested p26splussubstrate_XL.xlsx (PRIDE PXD064932): the import succeeded with 190 cross-links, while 107 rows failed — which seems reasonable, since entries like UCHL5_C88A and Ubiquitin_* look like construct/mutant labels rather than valid protein IDs.

From a code-review perspective nothing stood out as problematic, but I only had time for a high-level pass rather than a deep dive into every edge case.

Issue #184, opened 2026-01-12
Integrate AlphaFold DB search and load into a step


Elena-kal:
## Issue description
Introduce a new workflow step that allows users to provide a UniProt ID. Based on this ID, the system retrieves the corresponding predicted protein structure from the AlphaFold Database and stores the associated structural data. The stored data must persist across application runs and be suitable for long-term retention so that repeated retrievals are not required.

## Acceptance Criteria
- [ ] A user can input a valid UniProt ID as part of the workflow.
- [ ] system automatically fetches the predicted protein structure data for the given UniProt ID from the AlphaFold Database
- [ ] data is stored persistently and remains available across multiple runs
- [ ] If the same UniProt ID is provided again, the system reuses the previously stored data instead of downloading it again.
- [ ] Invalid or unknown UniProt IDs are handled gracefully with a clear error message.

PR #216, opened 2026-01-30
187 protein structures in settings

### LINKED ISSUE #187: Upload and manage predicted protein structures via settings
## Issue description
Users can upload predicted protein structures via the settings. During the upload, the user provides all required files for a protein structure prediction and enters the associated metadata (CIF file of the predicted structure, a PDB file of the predicted structure, a JSON file containing per-residue confidence scores, and a JSON file containing the predicted aligned error). 

Uploaded protein structure predictions are stored persistently and can be reused across sessions. The settings view also provides an overview of all previously uploaded protein structure predictions. From this overview, users can delete uploaded predictions if they are no longer needed.

## Acceptance Criteria
- [ ] The settings include an option to upload predicted protein structures.
- [ ] The upload requires the following files (for now): CIF structure file, PDB structure file, confidence JSON file, predicted aligned error JSON file.
- [ ] The user must provide metadata for each uploaded protein structure.
- [ ] The system validates that all required files are present and rejects incomplete uploads with a clear error message.
- [ ] The user can view a list of all previously uploaded protein structure predictions in the settings.
- [ ] The user can delete an uploaded protein structure prediction.
- [ ] The system prevents accidental duplicate uploads of the same protein structure or clearly indicates when a structure already exists.

Elena-kal:
## Description
fixes #187 
The user can now add own protein structure predictions in settings. They can delete already uploaded protein structures. They cannot enter protein structures with an EntryID that already exsists among the uploaded protein structures. When adding a new protein structure with the Alphafold DB API over a node in a workflow with an EntryId (in this case the same as the Protein ID) that already exsists, the exsisting will be overwritten and the user is notified with a warning. 

## Changes
Changes are made in frontend and backend. Most importantly is the logic in [views_settings.py](backend/main/views_settings.py) and [protein-structure-upload.tsx](frontend/src/components/app/settings/other-settings/protein-structure-upload.tsx). 

## Testing

1. Maybe add one or two protein structures in steps and check whether they show up in settings.
2. Try to add the same structure twice in the steps and check whether the warning shows up.
3. Try deleting structures in settings and check whether all the files are actually deleted by going into the folder userdata/external_data/alphafold.
4. Add your own structure (for the files I always used ones from alphafold because I did not have my own protein structure predictions at hand because I don't usually predict protein structures, but if you do go ahead and use them :))).
5. Try adding a structure in settings with an Entry ID that already exsists. (You should get an error, because the use should not do that.)

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**


backend/main/urls.py
```diff
@@ -61,6 +61,21 @@
...
+    ),
+    path(
+        "api/upload_prot_structure",
+        views_settings.upload_prot_structure,
+        name="upload_prot_structure",
+    ),
+    path(
+        "api/prot_structure_delete",
+        views_settings.prot_structure_delete,
+        name="prot_structure_delete",
```
tE3m: The paths, views and methods should probably be named consistently (in this case `delete_prot_structure`)

backend/main/views_settings.py
tE3m, line 1: I had issues with a `alphafold_metadata.csv` that was already present in my user_data directory where the column names were wrong/different. No idea where that was from. But the code doesn't create the file if it's not present, so that should be fixed
edit: I think I already checked out #184 at some point and that's where it came from

backend/main/views_settings.py
```diff
@@ -221,6 +228,165 @@ def save_ptm_settings(request, default_file_stem: str = DEFAULT_PTM_SETTINGS_FIL
...
+# <--- Protein Structure Predictions --->
+
+AF_DICT_PATH = EXTERNAL_DATA_PATH / "alphafold"
+
+
+def get_prot_structure(request):
+    metadata_csv = AF_DICT_PATH / "alphafold_metadata.csv"
+    df = pandas.read_csv(metadata_csv)
+
+    df_infos = df.rename(
```
tE3m, line 255: It'd probably be a good idea to filter the df to only the columns that should actually be output (as I said before, my file had more columns which led to errors with NaN values in the JSON parser)
edit: otoh, if we don't ever expect any changes to the data structure, this would be fine.

backend/main/views_settings.py
```diff
@@ -221,6 +228,165 @@ def save_ptm_settings(request, default_file_stem: str = DEFAULT_PTM_SETTINGS_FIL
...
+            af_path.mkdir(parents=True, exist_ok=True)
+
+        for file_name in [cif_file, confidence, pae, fasta_file]:
+            source_dir = settings.FILE_UPLOAD_TEMP_DIR / file_name
+            success, message = copy_file_to_directory(source_dir, af_path)
+
+        # add row to metadata csv
+        AF_DICT_PATH.mkdir(parents=True, exist_ok=True)
+        metadata_csv = AF_DICT_PATH / "alphafold_metadata.csv"
+        df = pandas.read_csv(metadata_csv)
```
tE3m: Same as with reading, this errors out if the file doesn't exist

backend/main/views_settings.py
```diff
@@ -221,6 +228,165 @@ def save_ptm_settings(request, default_file_stem: str = DEFAULT_PTM_SETTINGS_FIL
...
+        if af_path.exists():
+            return JsonResponse(
+                {"success": False, "message": "Entry ID is not unique."}, status=405
+            )
+        else:
+            af_path.mkdir(parents=True, exist_ok=True)
+
+        for file_name in [cif_file, confidence, pae, fasta_file]:
+            source_dir = settings.FILE_UPLOAD_TEMP_DIR / file_name
+            success, message = copy_file_to_directory(source_dir, af_path)
```
tE3m, line 292: `success` is never used. I guess this should be used for error handling? If not, please replace it with `_`

frontend/src/components/app/settings/other-settings/protein-structure-upload.tsx
```diff
@@ -0,0 +1,275 @@
...
+              type: "text",
+              name: "gene",
+              label: "Gene Name (required):",
+              isVisible: true,
+            },
+            {
+              type: "file",
+              name: "cif_file",
+              label: "CIF file (required):",
+              isVisible: true,
```
tE3m, line 210: For file inputs, I actually implemented the ability to specify the file endings that should be allowed (back in #127). Here, it would look like this:
```suggestion
              type: "file",
              name: "cif_file",
              label: "CIF file (required):",
              isVisible: true,
              accept: ".cif",
```

frontend/src/components/app/settings/other-settings/protein-structure-upload.tsx
```diff
@@ -0,0 +1,275 @@
...
+            },
+            {
+              type: "file",
+              name: "confidence_file",
+              label: "Confidence JSON file (required):",
+              isVisible: true,
```
tE3m, line 217:
```suggestion
              isVisible: true,
              accept: ".json",
```

frontend/src/components/app/settings/other-settings/protein-structure-upload.tsx
```diff
@@ -0,0 +1,275 @@
...
+            },
+            {
+              type: "file",
+              name: "pae_file",
+              label: "Predicted Aligned Error JSON file (required):",
+              isVisible: true,
```
tE3m, line 224:
```suggestion
              isVisible: true,
              accept: ".json",
```

frontend/src/components/app/settings/other-settings/protein-structure-upload.tsx
```diff
@@ -0,0 +1,275 @@
...
+            },
+            {
+              type: "file",
+              name: "fasta_file",
+              label: "Sequence FASTA file (required):",
+              isVisible: true,
```
tE3m, line 231:
```suggestion
              isVisible: true,
              accept: ".fasta, .fa",
```

frontend/src/components/app/settings/other-settings/protein-structure-upload.tsx
```diff
@@ -0,0 +1,275 @@
...
+            uniprot_id={ps.uniprot_id}
+            date_modified={ps.date_modified}
+            gene={ps.gene}
+            af_version={ps.af_version}
+            handleDelete={() => {
+              onDeleteProtStructure(ps.entry_id);
+            }}
+          />
+        ))}
+      </ProtStructureList>
```
tE3m: there should probably be a little info text here in case there aren't any predictions yet

backend/tests/protzilla/importing/test_alphafold_protein_structure_load.py
```diff
@@ -101,13 +101,15 @@ def test_fetch_alphafold_returned_keys(tmp_path, monkeypatch):
...
 
     out = af.fetch_alphafold_protein_structure("Q8WP00", persist_uploads=True)
-    assert set(out.keys()) == {
-        "metadata_df",
-        "cif_df",
-        "pae_df",
-        "plddt_df",
-        "sequence_df",
-    }
+    assert set(out.keys()).issuperset(
```
tE3m: Is there any reason not to be strict in this test? From what I can see, the `messages` attribute should always be set, so the key should always be present

backend/main/views_helper.py
```diff
@@ -184,3 +186,86 @@ def load_yaml_from_file(path: Path) -> str:
...
+
+    # Build error message
+    error_messages = []
+    if missing_files:
+        error_messages.append(f"Missing files: {', '.join(missing_files)}")
+    if invalid_files:
+        error_messages.append(f"Invalid file format: {', '.join(invalid_files)}")
+
+    if error_messages:
+        msg = " | ".join(error_messages)
```
tE3m: why do we concatenate the error messages? Usual protzilla style is a list of strings, if I'm not mistaken

tE3m (changes requested):

frontend/src/components/app/settings/other-settings/protein-structure-upload.tsx
Elena-kal: Thanks, that is really helpful!! saves me from using the validate input function that I had to implement just for this. 
Edit: Implemented this now and it is also so much prettier because it does not even let you choose the wrong file types. Thanks!

backend/main/views_settings.py
Elena-kal: I think we would have to touch the upload logic anyways if we wanted to change the data structure but I still changed it so only the columns that will be output are read because I think that is reasonable behavior.

backend/main/views_settings.py
Elena-kal: Thanks for pointing this out. The problem with this was also that even if one upload failed, if the next one did not we just ignored the failure and did not throw any errors. I changed it now and for that we need success.

frontend/src/components/app/settings/other-settings/protein-structure-upload.tsx
tE3m: One caveat with this: it's the browser's/file-picker's responsibility to handle the `accept` parameter of the file upload field. E.g. for me, I still have the option to ignore this and switch to all files being displayed. And the accept-field is of course only frontend-code, so if one were to simply use the route with a POST request from outside of our app, anything goes. While I feel like this should be enough for well-behaved, non-malicious users, it's still pretty easy to upload files that aren't what we expect. OTOH, we don't actually check the file content anyway, so idk if that's a big deal.

backend/main/views_settings.py
tE3m: I see that you changed this in the upload-method, is there a reason not to do so in the get-method as well? I get that it works, but maybe defining the expected columns as a constant and using them the same way here would be more consistent

tE3m (commented):
I'm happy with the changes, just some comments that don't necessarily need to be changed; up to you.

NeleRiediger (commented):
While performing the tests you suggested, I couldn't even import a protein structure in the steps. I always got the error message "An error occured while validating inputs or outputs: Output validation failed: missing output cif_df in outputs. Please check your parameters." for some different protein ids. However that is most likely a user issue one my side :)
Otherwise the code looks good to me.

PR #234, opened 2026-02-03
feat: add new importing step that loads data from disk into dataframe…

### LINKED ISSUE #230: New importing step for protein structure predictions that loads data from disk into dataframes
## Issue description
There should be another importing step where the user can choose a prediction that is loaded from disk into dataframes.

## Acceptance Criteria
- [ ] a new step in importing
- [ ] one can choose an entryID of which the data should be loaded into dataframes
- [ ] all files on disk can be chosen in the step
- [ ] for the protein structure chosen all data is loaded into dataframes

Elena-kal:
## Description
fixes #230 
Adds a new importing step where you can choose between the different protein structure predictions on disk (entryIds given) which one you want to load into the run. All data regarding the predictions are then loaded into dataframes in the same format one would have if one loaded the data directly from alphafold db via the alphafold load step.

## Changes
New step in forms [importing.py](backend/protzilla/methods/importing.py). The logic can be found in [alphafold_protein_structure_load.py](backend/protzilla/importing/alphafold_protein_structure_load.py). 

## Testing
Create a run, make sure to have some predictions in user data (either by loading them via the step or uploading them in settings) and choose any of them to load the dataframes. You should be able to see them in the tables tab. 
(Feel free to also experiment with what happens when you write new entries into the csv when there are no folders or try to break it in anyway you think is reasonable. Although that is not too necessary as I have tested that to some extend and I think we can assume users are not malicious and actively try to break things.)

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**


backend/protzilla/methods/importing.py
```diff
@@ -462,3 +465,33 @@ def create_form(self):
         )
 
     calc_method = staticmethod(crosslinking_import)
+
+
+class ImportStructurePredictionsFromDisk(ImportingStep):
+    display_name = "Structure Predictions Import from Disk"
```
AnnaPolensky: Per step-instance you can only import the structure prediction of one protein, right? In that case, I would maybe call it "Structure Prediction Import from Disk" ?

backend/protzilla/methods/importing.py
```diff
@@ -462,3 +465,33 @@ def create_form(self):
...
+        "metadata_df",
+        "cif_df",
+        "pae_df",
+        "plddt_df",
+        "sequence_df",
+    ]
+
+    def create_form(self):
+        return Form(
+            label="Structure Predictions Import from Disk",
```
AnnaPolensky, line 491: Same as above

backend/protzilla/importing/alphafold_protein_structure_load.py
```diff
@@ -198,6 +198,146 @@ def handle_alphafold_files(
     }
 
 
+def get_all_available_entry_ids() -> list:
```
AnnaPolensky: Docstring?

backend/protzilla/importing/alphafold_protein_structure_load.py
```diff
@@ -198,6 +198,146 @@ def handle_alphafold_files(
...
+    :param entry_id: entryID of the uploaded protein structure
+    :return: A dictionary containing DataFrames for metadata, CIF, PAE, pLDDT, and sequence data
+    """
+    messages: list[dict[str, Any]] = []
+
+    meta_dir = paths.EXTERNAL_DATA_PATH / "alphafold"
+    metadata_csv = meta_dir / "alphafold_metadata.csv"
+
+    if not metadata_csv.exists():
+        msg = f"AlphaFold metadata CSV not found: {metadata_csv}"
```
AnnaPolensky: Do we want to put AlphaFold into the message, since this could also be a user-generated prediction?
(same for l. 239)

backend/protzilla/importing/alphafold_protein_structure_load.py
```diff
@@ -198,6 +198,146 @@ def handle_alphafold_files(
...
+        msg = f"No JSON files (PAE/pLDDT) found in {prot_dir} for entry '{entry_id}'"
+        logger.error(msg)
+        raise FileNotFoundError(msg)
+
+    pae_df = None
+    plddt_df = None
+
+    try:
+        if len(json_files) == 1:
+            raise RuntimeError()
```
AnnaPolensky, line 301: Can we make this error a little bit more descriptive?

backend/protzilla/importing/alphafold_protein_structure_load.py
```diff
@@ -198,6 +198,146 @@ def handle_alphafold_files(
...
+
+    if (
+        _check_df(cif_df)
+        and _check_df(pae_df)
+        and _check_df(plddt_df)
+        and _check_df(sequence_df)
+    ):
+        success_msg = f"Successfully loaded AlphaFold data for entry '{entry_id}'"
+        logger.info(success_msg)
+        messages.append(dict(level=logging.INFO, msg=success_msg))
```
AnnaPolensky: Every time we miss a file, we raise an error. Do you know whether that is consistent with the rest of the codebase or whether we would just want to add msg to messages (like we add the success_msg in this line here)?

backend/protzilla/importing/alphafold_protein_structure_load.py
```diff
@@ -198,6 +198,146 @@ def handle_alphafold_files(
...
+    except Exception as e:
+        msg = f"Failed to read JSON files in {prot_dir}: {e}"
+        logger.exception(msg)
+        raise RuntimeError(msg) from e
+
+    def _check_df(df: Any) -> bool:
+        return df is not None
+
+    if (
+        _check_df(cif_df)
```
AnnaPolensky: At first, I thought this function checked whether the df had the right columns or something. Maybe we could rename it as "_check_df_not_empty". Or I would also be fine with just replacing the function calls with the function code as it is really not a lot of code we are extracting.

backend/tests/protzilla/importing/test_alphafold_protein_structure_load.py
```diff
@@ -166,3 +170,123 @@ def test_fetch_alphafold_dfs_exist(tmp_path, monkeypatch):
...
+
+    assert isinstance(out["pae_df"], pd.DataFrame)
+    assert not out["pae_df"].empty
+
+    assert isinstance(out["plddt_df"], pd.DataFrame)
+    assert not out["plddt_df"].empty
+
+    assert isinstance(out["sequence_df"], pd.DataFrame)
+    assert not out["sequence_df"].empty
+
```
AnnaPolensky, line 290: Would it make sense not only to check that the dataframes aren't empty, but also that they do contain the right things?

AnnaPolensky (commented):
Works for me and looks good overall, although the tests are failing right now.

backend/protzilla/importing/alphafold_protein_structure_load.py
```diff
@@ -198,6 +198,146 @@ def handle_alphafold_files(
     }
 
 
+def get_all_available_entry_ids() -> list:
+    meta_dir = paths.EXTERNAL_DATA_PATH / "alphafold"
+    metadata_csv = meta_dir / "alphafold_metadata.csv"
```
tE3m: I feel like these paths appear often enough to add them to constants/paths.py

backend/protzilla/importing/alphafold_protein_structure_load.py
tE3m, line 1: btw, I know that this isn't part of this PR, but in `read_alphafold_mmcif`, the if-branch checking if the returned table is `None` does nothing because the method always returns a table - if it's empty, the columns get defined without any rows.

backend/protzilla/importing/alphafold_protein_structure_load.py
tE3m:
```suggestion
def get_all_available_entry_ids() -> list[str]:
```

backend/protzilla/importing/alphafold_protein_structure_load.py
```diff
@@ -198,6 +198,146 @@ def handle_alphafold_files(
...
+
+
+def get_prot_structure_dfs(entry_id: str) -> dict[str, Any]:
+    """
+    Writes data from disk of a specific entry ID into dataframes.
+
+    :param entry_id: entryID of the uploaded protein structure
+    :return: A dictionary containing DataFrames for metadata, CIF, PAE, pLDDT, and sequence data
+    """
+    messages: list[dict[str, Any]] = []
```
tE3m: `Any` is not a good type hint, especially if we know the types that can occur (logging levels are integers, messages are strings)
```suggestion
    messages: list[dict[str, str | int]] = []
```

backend/protzilla/importing/alphafold_protein_structure_load.py
tE3m: The likelihood of the prediction being done through AlphaFold (even if it's user-generated) is high, but it's true that there are also other tools. OTOH, the files are called alphafold as well, so..

backend/protzilla/importing/alphafold_protein_structure_load.py
tE3m: And while we're at it, there should probably be a method for retrieving the metadata csv, as to not re-write that everytime it's used and do the file handling over and over

backend/protzilla/importing/alphafold_protein_structure_load.py
```diff
@@ -198,6 +198,146 @@ def handle_alphafold_files(
...
+        raise FileNotFoundError(msg)
+
+    # get cif file
+    cif_files = list(prot_dir.glob("*.cif"))
+    if not cif_files:
+        msg = f"No CIF file found in {prot_dir} for entry '{entry_id}'"
+        logger.error(msg)
+        raise FileNotFoundError(msg)
+
+    cif_file = cif_files[0]
```
tE3m, line 259: Is there a specific reason to pick the first one? Do we assume that there will only ever be one? There should probably be a comment here. Maybe also a size check and warning in case there are ignored cif files

backend/protzilla/importing/alphafold_protein_structure_load.py
```diff
@@ -198,6 +198,146 @@ def handle_alphafold_files(
...
+                "predicted_aligned_error" in json2.columns
+                and "residueNumber" in json1.columns
+            ):
+                pae_df = json2
+                plddt_df = json1
+            else:
+                # Fallback: assign and warn
+                pae_df = json1
+                plddt_df = json2
+                warn = f"Could not detect PAE/pLDDT in JSON files for entry '{entry_id}'; files will be returned as read."
```
tE3m: I actually got this error while experimenting. I don't think this conveys what's going on; the user should be told which file was assigned to which df.

backend/protzilla/importing/alphafold_protein_structure_load.py
```diff
@@ -198,6 +198,146 @@ def handle_alphafold_files(
...
+        messages.append(dict(level=logging.INFO, msg=success_msg))
+
+    return {
+        "metadata_df": metadata_df,
+        "cif_df": cif_df,
+        "pae_df": pae_df,
+        "plddt_df": plddt_df,
+        "sequence_df": sequence_df,
+        "messages": messages,
+    }
```
tE3m: I think this reads a lot better and simpler if we construct the dict earlier and iterate over it
```suggestion
    df_dict = {
        "metadata_df": metadata_df,
        "cif_df": cif_df,
        "pae_df": pae_df,
        "plddt_df": plddt_df,
        "sequence_df": sequence_df,
    }

    if not any(df.empty for df in df_dict.values()):
        success_msg = f"Successfully loaded AlphaFold data for entry '{entry_id}'"
        logger.info(success_msg)
        messages.append(dict(level=logging.INFO, msg=success_msg))

    df_dict["messages"] = messages

    return df_dict
```

backend/tests/protzilla/importing/test_alphafold_protein_structure_load.py
```diff
@@ -1,14 +1,18 @@
...
+import logging
 
 
 from backend.protzilla.importing.alphafold_protein_structure_load import (
     fetch_alphafold_protein_structure,
     to_fasta,
     read_alphafold_mmcif,
+    get_all_available_entry_ids,
+    get_prot_structure_dfs,
+    paths,
```
tE3m: This is a bit odd, isn't it? The paths are an import there as well. Should import them directly from constants/paths.py

backend/protzilla/importing/alphafold_protein_structure_load.py
tE3m: Also, is there a reason to only check for success? There's not even a warning if any of the dataframes are empty

tE3m (changes requested):

backend/protzilla/importing/alphafold_protein_structure_load.py
Elena-kal: I think that is what is intended for now and if the user really wants to upload their own data then they will have to format it the way alphafold would anyway so I guess it would be like turning their own predictions into predictions that could have been done by alphafold. I'd keep the naming for now and if we decide to change it I would change it everywhere and not just here.

backend/protzilla/importing/alphafold_protein_structure_load.py
Elena-kal: I do not think it is exactly consistent because in many places of the codebase we do not even check for things that could go wrong. However, I would still argue it is the most reasonable behavior. If we do not raise an error but simply append to messages we keep going with our code but with faulty parts, possibly breaking even more and simply causing more error messages to be appended and in the end returning an output that we cannot work with. Much better to throw an error.

tE3m (approved):
Changes look great!

PR #236, opened 2026-02-04
Multiple organism id input for crosslinking import 

### LINKED ISSUE #235: Multiple organism id input for crosslinking import
## Issue description
Since a crosslinking file that is imported can contain proteins from more than just one organism, the user should be able to input multiple organism ids. Those should be ordered after priority, so that the first uniprot lookups is done with the first organism id. If after that there are proteins without a lookup result, those should be looked up with the second organism id and so on and so forth. 

## Acceptance Criteria
- [x] The user is able to input multiple organisms ids 
- [x] The user is able to prioritize those organism ids 
- [x] It is obvious to the user, that there is a prioritization 
- [x] The calc method takes a sorted list of organism ids 
- [x] Every organism id of this list is validated  
- [x] If the validation fails for any organism id of the list, the whole process is terminated with an error 
- [x] Error message when terminating because of a wrong organism id, shows which organism id was the problem 
- [x] The lookup with an organism id is performed in several steps, starting with the first organism id and then a second lookup with the second organism id, and all proteins that don't have a result yet

NeleRiediger:
## Description
fixes #235 
Changes follow exactly what is described in this issue. 

## Changes
backend/protzilla/methods/importing.py ➞ some changes to the organism id input field (mostly added descriptions) 
backend/protzilla/importing/crosslinking_inport.py ➞ changes to the logic so that a list of organism ids can be processed 

## Testing
Enter several different lists of organism ids and try to import both file types with those. Try some intentionally wrong inputs and check if the error messages you see correspond to what you did wrong.

Maybe try to import a csm file with the organism id 9606 and/ or 10090 and see what difference that makes in the failed proteins, that are shown in the warning after the import. 

(The acceptance criteria in the issue #235 might also give some ideas on what to test) 

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**

### DIFF
--- a/backend/protzilla/importing/crosslinking_import.py
+++ b/backend/protzilla/importing/crosslinking_import.py
@@ -287,12 +287,10 @@ def uniprot_lookup(
     :rtype: None
     """
     if mode == ProteinDesignationLookupMode.id_to_gene_name.value:
-        error = ProteinLookupError.NO_GENE_NAME_FOUND.value
         field_of_existing_data = "accession"
         extra_query = None
         extra_fields = None
     elif mode == ProteinDesignationLookupMode.gene_name_to_id.value:
-        error = ProteinLookupError.NO_PROTEIN_ID_FOUND.value
         field_of_existing_data = "gene_exact"
         extra_query = f"organism_id:{organism_id} AND reviewed:true"
         extra_fields = "gene_names"
@@ -315,10 +313,6 @@ def uniprot_lookup(
         process_uniprot_response(
             response=response, results=results, input_data=batch, mode=mode
         )
-
-    for data in input_data:
-        if data not in results:
-            results[data] = (False, None, error)
 
 
 def get_gene_name_from_protein_ids(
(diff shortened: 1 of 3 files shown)


backend/protzilla/importing/crosslinking_import.py
```diff
@@ -643,35 +663,52 @@ def normalize_crosslinking_df(df: pd.DataFrame) -> pd.DataFrame:
...
+
+    organism_ids_for_request = ",".join(organism_ids_list)
+    url = f"https://eutils.ncbi.nlm.nih.gov/entrez/eutils/esummary.fcgi?db=taxonomy&id={organism_ids_for_request}&retmode=json"
+    try:
+        response = requests.get(url, timeout=15)
+        if response.status_code != 200:
+            return False, None, None
+        data = response.json()
+    except Exception:
+        return False, None, None
```
AnnaPolensky: I'm unsure how to test this, but based on this code it looks like we are giving the same error message when the input provided was wrong and when we had a problem getting an answer for the request (e.g. because there was a timeout). I think it would make sense to provide a message that it might not have been the user's fault?

AnnaPolensky (commented):
Code looks good :)

Two things I noted while testing:
- for the xlsx: If I enter an invalid ID like -1, no error shows up. This confused me at first. Maybe we could make it clearer that the IDs are not used in the xlsx?
- Would it be nice to show for how many rows which organism ID was used (in the message) or to add the used organism ID to the df, if there is any use for this later on? Just a thought, but it might be completely unnecessary, so definitely ask Chris first whether this makes sense!

backend/protzilla/importing/crosslinking_import.py
```diff
@@ -709,44 +750,47 @@ def aggregate_failed_proteins_for_display(failed_df: pd.DataFrame) -> str:
...
+                level=logging.ERROR,
+                msg=msg,
+                trace=trace,
+            )
+        ],
+    )
+
+
+def crosslinking_import(file_path: Path, organism_ids: str) -> dict:
     file_type = file_path.suffix
```
3dot141592: Just a little detail, but file extensions are handled case‑sensitively here (.CSV / .XLSX will be rejected). Maybe use
```suggestion
    file_type = file_path.suffix.lower()
```
instead.

3dot141592 (approved):
Apart from this small detail, all tested cases worked well as far as my understanding of the expected behavior goes. The code itself looks all right to me.

PR #238, opened 2026-02-04
Add server-side pagination for tables

### LINKED ISSUE #218: Minmise data traffic
## Issue description
Every time you navigate to any step, the entire output table of that step gets sent to the front-end. This results in horrible performance (e.g. 40MiB of data sent on every navigation!) You can verify this on Firefox with the Network Tab in the Developer Tools. On local hosts, this is still acceptable. As soon as PROTzilla is used in a networked environment with a server hosting the application, performance will likely be unacceptable.

## Possible fixes
- Only send table when the table view is selected
- Only send parts of table unless we explicitly request more from the back-end (i.e. only ever send the parts actually displayed by the front-end)

jorisfu:
## Description
fixes #218 
PROTzilla now only sends table data to the front-end that is actually displayed. New API calls were added for this and the front-end was adjusted as necessary

## Changes
- New API call: `get_current_step_output_labels` sends the list of output labels of the current step (the actual labels used by the back-end to identify outputs of a step) mapping to their corresponding "display names", aka how the table names will be displayed to the user
- New API call: `get_current_step_table_data` retrieves the table data for a given output label satisfied by the currently selected step. Allows the caller to delimit the data sent via an index range
- Removed API call: `get_step_table` previously sent the entirety of all tables of the current step to the front-end in bulk. This has been made obsolete by the other API endpoints.
- Updated `run-screen` to only fetch the output labels when fetching data on step navigation/loading
- Updated `data-table` to accept a run name and table label to then dynamically pull the data from the back-end when paging through the table
- Updated `CSVButton` similarly, now downloads the entire table through a separate request, also indicates loading state when downloading
- Updated `switch.tsx` to fix the `[Object object]` issues (same thing Robin did but I got annoyed by it so it's included here)

## Testing
Use any existing run and make sure the tables all show up as expected. You can double check whether or not the tables are the same by comparing the output prior to this PR (slicing _should_ not change any ordering). You can also measure the performance benefit on large runs (e.g. AML paper data): steps should load noticeably faster. Still too slow due to the plots, but definitely faster than previously

## Note
When merging, please squash the commits (they're mostly chaotic pipeline fixes after the first major commit)

## PR checklist
**Development**
- [ ] If necessary, I have created / updated tests. (API calls do not get tested currently)
 
**Mergeability**
 
**Code review**


frontend/src/components/core/data-table/data-table.stories.tsx
```diff
@@ -360,9 +359,10 @@ const exampleData: GridRowsProp = [
     id: 49,
   },
 ];
+*/
```
NeleRiediger: Is this supposed to be deleted?

frontend/src/components/app/run-screen/run-screen.tsx
```diff
@@ -205,29 +205,27 @@ export const RunScreen: React.FC = () => {
...
+            name: output_info.display_name,
           }))}
         />
       ) : (
         <SectionTitle
           baseComponent={"h4"}
-          description={
-            "No data table available for this step (yet). With large datasets it may take a while for tables to be displayed."
-          }
+          description={"This step does not provide any tables as output"}
```
NeleRiediger: Really just a tiny little thing, but perhaps add a period at the end of that comment. Somehow this bugs me 😅

NeleRiediger (commented):
When I first tested I only ever got the "This step does not provide any tables as output" everywhere. But then I double checked whether those tables should have been there on the dev branch and when I changed back to this branch everything worked as expected. So I guess that was just some weird issue that is not actually an issue. 

Nonetheless I have to admit that I didn't really notice any improvement in the peptide import for example (tested that one because I had previously noticed that it was quite slow). Probably this is just a bad example and actual improvement is just noticeable on even larger and slower steps. 
So apart from my really tiny comments everything looks good to me.

frontend/src/components/core/data-table/data-table.stories.tsx
jorisfu: yeah, fixed it

frontend/src/components/app/run-screen/run-screen.tsx
jorisfu: Fixed

jorisfu:
> Nonetheless I have to admit that I didn't really notice any improvement in the peptide import for example

The improvement is relatively small when calculating new steps, as the calculation itself takes far longer than sending the entire output table would. What this mostly addresses is navigation between steps already calculated. You should be able to notice a significant speedup in loading times there (e.g. with the AML paper dataset)

NeleRiediger (approved):

3dot141592 (approved):
I tried this on the AML paper dataset in the frontend. Before this change, navigating steps would repeatedly load the full table payloads (~25 MB). With the new paging, only small slices are fetched :). 

Code looks good overall.

3dot141592:
Maybe we could introduce some lightweight speed/performance tests so future changes dont accidentally reintroduce large payload bloat. (yes, I **love** tests).

jorisfu:
> Maybe we could introduce some lightweight speed/performance tests so future changes dont accidentally reintroduce large payload bloat. (yes, I **love** tests).

Generally agree, probably hard to test in practice though. We'd need some kind of selenium workflow in order to test that kind of front-end interaction reliably.

Merging and closing.

PR #242, opened 2026-02-06
Adjust heatmap / clustergram

### LINKED ISSUE #217: Fix Hierarchical Clustering
## Issue description
Hierarchical Clustering has some issues. Will update this Issue's TODOs as I identify and fix them. Closely related to #201 and I'm working directly on `test-branch-for-paper` for this

## The Problems
- [ ] Saving and reloading a run with Hierarchical Agglomerative Clustering (HAC) doenst work as the `model` data is saved as `!!python/object:sklearn.cluster._agglomerative.AgglomerativeClustering` and there is no constructor to reconstruct the data from that
- [ ] HAC does not produce any plots. Getting the HAC step output from sklearn converted to something that the existing Clustergram plot class supports may be a major pain to implement. Sklearn itself only supports matplotlib dendrograms as outputs
- [ ] The "Plot: Clustergram" step and HAC cannot take input from a t-test
- [ ] The "Plot: Clustergram" step and HAC cannot deal with NaNs

jorisfu:
## Description
Addresses #217 

Changes some things with the clustergram step to more closely match the AML paper

## Changes
- Allow for custom colour scales (low and high colours including respective z-score boundaries)
- Asymmetric colour scales will always retain 0 mapping to white
- Allow for custom user-provided labels for the heatmap colour legend
- Colour legend ticks always show minimum z, maximum z, and 0.
- Change default colours to match AML paper

## Still TODO
- Verify major clusters in our branch matching those in the AML paper supplementary data set

## Testing
Create a Clustergram (e.g. with the AML paper data) and experiment with the new settings

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**


backend/protzilla/data_analysis/plots.py
Elena-kal, line 230: doc string should be updated

backend/protzilla/data_analysis/plots.py
```diff
@@ -253,7 +259,14 @@ def clustergram_plot(
             row_colors = None
             color_label_dict = None
 
-        # TODO: Would be nice to actually center values at the z-score of 0
+        if use_custom_colour_scale:
+            custom_colour_scale = (
+                (heatmap_low_colour_limit, heatmap_low_colour),
+                (heatmap_high_colour_limit, heatmap_high_colour),
```
Elena-kal: how sure are we that we use british english in our code base? I always thought it was american in the code and british for everything the user reads and sees. 

Image

backend/protzilla/utilities/clustergram.py
```diff
@@ -528,21 +536,44 @@ def figure(self, computed_traces=None):
...
+                high_colour = self._custom_colour_scale[1][1]
+
+            else:
+                zmin = np.min(heat_data)
+                zmax = np.max(heat_data)
+                low_colour = HEATMAP_LOW_COLOUR
+                high_colour = HEATMAP_HIGH_COLOUR
+
+            # Lerp to keep zero centered
+            midpoint = (0 - zmin) / (zmax - zmin)
```
Elena-kal: How sure are we that it is always zmax != zmin?

backend/protzilla/utilities/clustergram.py
```diff
@@ -528,21 +536,44 @@ def figure(self, computed_traces=None):
...
+
+            # If range of values is only positive or only negative, lerp doesn't work.
+            # In this case we reset. Shouldn't happen for most normalised data sets
+            # which are usually plotted with heatmaps
+            if midpoint < 0 or midpoint > 1.0:
+                midpoint = 0.5
```
Elena-kal: Why 0.5?

Elena-kal (approved):
It works fine. The plot itself does not look too pretty but I think that is simply the nature of clustergrams. I wrote a couple of comments but overall I think the code is good.

backend/protzilla/utilities/clustergram.py
```diff
@@ -528,21 +536,44 @@ def figure(self, computed_traces=None):
             heat_data = self._data
 
             # symmetrize the heatmap about zero, if necessary
-            if self._center_values:
-                heat_data = np.subtract(heat_data, np.mean(heat_data))
+            # if self._center_values:
+            #     heat_data = np.subtract(heat_data, np.mean(heat_data))
```
3dot141592: Question: Should this centering remain removed permanently, or was it only temporary for the new color scheme?

3dot141592 (commented):
Overall looks good to me aside from the one comment. 

One remaining concern: hierarchical agglomerative clustering still does not accept NaN values. How do we want to handle this going forward? document/expect the user to pre-clean, or implement automatic imputation?

backend/protzilla/utilities/clustergram.py
```diff
@@ -23,6 +23,10 @@
 from sklearn.impute import SimpleImputer
 
 
+HEATMAP_LOW_COLOUR = "#0000FF"
+HEATMAP_HIGH_COLOUR = "#FF0000"
```
hendraet: Why not make this dependent on the min/max values of the `RdBu_r` color to make it more consistent with the usual plotly colorscale? (Might also make it slightly less ugly)

backend/protzilla/utilities/clustergram.py
hendraet: Would at least change the default values because executing the step as is definitely raises this ZeroDivisionError

hendraet: Would also add a sanity check that ensures that zmin is actually smaller than zmax. It doesn't throw an error, but it also doesn't look like it behaves correctly (e.g. heatmap range is labeled weirdly)

backend/protzilla/data_analysis/plots.py
```diff
@@ -253,7 +259,14 @@ def clustergram_plot(
             row_colors = None
             color_label_dict = None
 
-        # TODO: Would be nice to actually center values at the z-score of 0
+        if use_custom_colour_scale:
+            custom_colour_scale = (
+                (heatmap_low_colour_limit, heatmap_low_colour),
```
hendraet: Could need some sanity checking or at least a proper error message if user enters sth. that is not a valid color

backend/protzilla/utilities/clustergram.py
```diff
@@ -528,21 +536,44 @@ def figure(self, computed_traces=None):
...
+                midpoint = 0.5
 
             heatmap = go.Heatmap(
                 x=tickvals_col,
                 y=tickvals_row,
                 z=heat_data,
-                colorscale=self._color_map,
-                # TODO: This should be based on the text width of the labels,
-                # or at least passable by the user, so they can adjust it
+                colorscale=[[0, low_colour], [midpoint, "white"], [1, high_colour]],
```
hendraet: Nit: setting `midpoint` to white limits the customizability of the plot. I guess it might be a bit too much to also add a field to customize this variable, but since I thought about it, I just wanted to leave this here.

backend/protzilla/utilities/clustergram.py
```diff
@@ -528,21 +536,44 @@ def figure(self, computed_traces=None):
...
+                zmin=zmin,
+                zmax=zmax,
                 colorbar=dict(
-                    title="Heatmap Legend",
+                    title=self._heatmap_legend_title,
                     yanchor="bottom",
                     y=0.0,
                     len=1 / self.colorbar_count,
+                    tickmode="array",
+                    tickvals=(zmin, 0, zmax),
```
hendraet: Midpoint is not shown in legend if both values are positive/negative

hendraet (changes requested):
Works nicely, and I really like the added customization options. 
The only thing I'm missing is better error handling and corresponding test cases, i.e., what happens if zmin and zmax are the same or inverted and what if the user enters an invalid color.

backend/protzilla/data_analysis/plots.py
jorisfu: Fixed this by adding a custom colour picker, is cleaner anyway and doesn't require sanity checking down the line (since plotly can also take other values for colours apart from hex codes)

backend/protzilla/utilities/clustergram.py
jorisfu: mistake on my part, should not have been deleted.

backend/protzilla/utilities/clustergram.py
jorisfu: Fascinatingly, using min/max of RdBu doesn't look that good, because RdBu uses more than min/mid/max colours to interpolate:

mynd

Should I still change it to this? I adjusted it to these defaults to match the (ugly) colour scale in the AML paper

backend/protzilla/utilities/clustergram.py
jorisfu: Fair point, this could be addressed in the future. Helps people create more misleading plots as well :)

jorisfu:
@hendraet @Elena-kal I addressed every other comment y'all made with these commits. Since this changes a lot of what I previously did all over again and I also added a new front-end component (colour picker so we don't need to validate colours down the line), it'd be cool if you could re-review

backend/protzilla/utilities/clustergram.py
hendraet: I admit, it's a bit underwhelming (although I still like it slightly better than the original colors). But I totally get the notion to adapt it to the color scale in the AML paper, so I wouldn't complain about using these colors (since they are customizable anyways)

backend/protzilla/data_analysis/plots.py
hendraet: During review, I thought that a color picker would be ideal, but I deemed it too much work to propose it. I love that you did implement one anyways :blush:

backend/protzilla/utilities/clustergram.py
```diff
@@ -531,18 +540,53 @@ def figure(self, computed_traces=None):
...
+                zmin = np.min(heat_data)
+                zmax = np.max(heat_data)
+                low_color = HEATMAP_LOW_COLOR
+                high_color = HEATMAP_HIGH_COLOR
+
+                # Really only zmin == zmax can happen here
+                if not zmin < zmax:
+                    raise ValueError(
+                        "Data consists only of identical values. Not plotting."
+                    )
```
hendraet, line 564: Nit: move this check to the top, or at least before the trace calculation. Otherwise, you have to wait for the computation to finish, which might take some time, only to get an error afterwards.

hendraet: Maybe same with the color check above (although I'm not sure if that one is needed anymore)

hendraet (commented):
I like all the added improvements! Only a minor suggestion

hendraet (approved):

backend/protzilla/utilities/clustergram.py
jorisfu: I added the same check in `_compute_clustered_data`, just to make sure if calculation changes somehow we still have this one in place.

PR #245, opened 2026-02-08
232 crosslinker position validation

### LINKED ISSUE #232: crosslinker position validation
## Issue description
In crosslinking_df there are the columns Peptide_position1 and Peptide_position2. These positions might be outdated. We want to remove the columns from the dataframe and calculate the positions of crosslinkers based on the peptide and protein sequences. Furthermore, we want to duplicate rows if there are several possible positions.

## Acceptance Criteria
- [x] Removed Peptide_position1 and Peptide_position2 from crosslinking_cf and everywhere where they show up in the code.
- [x] Updated crosslinking_df documentation.
- [x] Check for all matches between the peptide sequence and protein sequence. 
- [x] If there is more than one match, duplicate the row in the dataframe.
- [x] If there is more than one match, give a warning to the user

AnnaPolensky:
## Description
fixes #232 
Removed the columns Peptide_position1 and Peptide_position2 from the crosslinking_df since they were not used and we don't want to use them since they might be outdated.
We match the peptide sequence where the crosslinker binds to the protein sequence to determine the 1-based amino acid position of the crosslinker. If there are several matches, we are duplicating the row in the dataframe and are giving a warning to the user.

## Changes
backend/protzilla/data_analysis/crosslinking_validation.py: Added function add_positions_of_amino_acid_where_crosslinker_bound_to_df, which contains most of the logic behind determining positions and duplicating rows. get_position_of_amino_acid_crosslinker_bound_to and get_distance_between_crosslinker_connected_amino_acids_in_alphafold were redundant afterwards.
Also added tests.

## Testing
I slightly changed one field of the p26splussubstrate_XL file. Now there is one crosslink with one peptide having 5 matches. Upload this file and check the plot and dataframe and check that there are 11 crosslinks instead of 7 for O43242. Feel free to change other Peptide sequences of the protein you validate as well, so that they are unmatchable/have more than one match.

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**


backend/protzilla/data_analysis/crosslinking_validation.py
```diff
@@ -101,63 +103,91 @@ def get_distance_between_two_amino_acids_in_angstrom(
...
+        all_position_combinations = [
+            (
+                pos1 + crosslinker_row.CL_position_within_peptide1 + 1,
+                pos2 + crosslinker_row.CL_position_within_peptide2 + 1,
+            )
+            for pos1 in peptide1_positions
+            for pos2 in peptide2_positions
+        ]
+        if not all_position_combinations:
+            msg = f"At least one of the peptide sequences ({peptide_sequence1}, {peptide_sequence2}) of crosslink entry {idx} was not found in the protein sequence. The entry was deleted."
```
Elena-kal: One could also tell the user which one of peptide sequence was not found but its not too necessary.

backend/protzilla/data_analysis/crosslinking_validation.py
```diff
@@ -101,63 +103,91 @@ def get_distance_between_two_amino_acids_in_angstrom(
...
+
+    if not rows_to_duplicate:
+        return crosslinking_df, messages
+    new_rows = []
+    for row_to_duplicate_idx, potential_positions in rows_to_duplicate.items():
+        for potential_cl_position1, potential_cl_position2 in potential_positions:
+            new_row = crosslinking_df.loc[row_to_duplicate_idx].copy()
+            new_row["crosslinker_position1"] = potential_cl_position1
+            new_row["crosslinker_position2"] = potential_cl_position2
+            new_rows.append(new_row)
```
Elena-kal, line 177: Couldn't you add the new row right here to the dataframe. Saves you the lines below, right?

backend/protzilla/data_analysis/crosslinking_validation.py
```diff
@@ -221,22 +260,37 @@ def check_crosslink(crosslink: pd.Series) -> pd.Series:
...
+                "crosslinker_position1": crosslink.crosslinker_position1,
+                "crosslinker_position2": crosslink.crosslinker_position2,
+            }
+        )
 
-    # adding the distance in alphafold and the result of the validation to all relevant crosslinks
-    all_crosslinks_df.loc[mask, ["alphafold_distance", "valid_crosslink"]] = (
-        relevant_crosslinks_df.apply(check_crosslink, axis=1)
+    # adding the distance in alphafold, the result of the validation and the crosslinker positions to all relevant crosslinks
+    new_colums = [
```
Elena-kal:
```suggestion
    new_columns = [
```

backend/protzilla/data_analysis/crosslinking_validation.py
```diff
@@ -221,22 +260,37 @@ def check_crosslink(crosslink: pd.Series) -> pd.Series:
...
+        "alphafold_distance",
+        "valid_crosslink",
+        "crosslinker_position1",
+        "crosslinker_position2",
+    ]
+    relevant_crosslinks_df[new_colums] = relevant_crosslinks_df.apply(
```
Elena-kal:
```suggestion
    relevant_crosslinks_df[new_columns] = relevant_crosslinks_df.apply(
```

backend/protzilla/data_analysis/crosslinking_validation.py
```diff
@@ -181,13 +211,12 @@ def validate_with_angstrom_deviation(
     protein to validate) of crosslinking_df and two more colums containing the distances in AlphaFold and wheter the crosslink matches the
```
Elena-kal:
```suggestion
    protein to validate) of crosslinking_df and two more colums containing the distances in AlphaFold and whether the crosslink matches the
```

backend/protzilla/data_analysis/crosslinking_validation.py
```diff
@@ -197,9 +226,19 @@ def validate_with_angstrom_deviation(
...
+    )
+
     def check_crosslink(crosslink: pd.Series) -> pd.Series:
-        distance = get_distance_between_crosslinker_connected_amino_acids_in_alphafold(
-            fasta_df, cif_df, crosslink
+        predicted_distance = get_distance_between_two_amino_acids_in_angstrom(
+            amino_acid_position1=crosslink.crosslinker_position1,
+            amino_acid_position2=crosslink.crosslinker_position2,
+            amino_acid_kind1=protein_sequence[crosslink.crosslinker_position1 - 1],
+            amino_acid_kind2=protein_sequence[crosslink.crosslinker_position2 - 1],
```
Elena-kal: Where is the difference between amino_acid_type and amino_acid_kind? You use both in this file. If they mean the same thing it should probably be consistent.

Elena-kal (approved):
Just a couple of small comments. Overall, it seems fine. It also worked just as you described.

backend/protzilla/data_analysis/crosslinking_validation.py
AnnaPolensky: I did this at the beginning, but I think it caused some issues (not sure anymore what exactly the problem was). However, concat also creates a new DataFrame, so it's more efficient to copy the whole DataFrame only once in the end.

backend/protzilla/data_analysis/crosslinking_validation.py
AnnaPolensky: There is no difference between them, I renamed them all to amino_acid_type.

backend/protzilla/data_analysis/crosslinking_validation.py
```diff
@@ -78,86 +80,117 @@ def get_distance_between_two_amino_acids_in_angstrom(
...
 
     return float(np.linalg.norm(pos2 - pos1))
 
 
-def get_position_of_amino_acid_crosslinker_bound_to(
-    protein_sequence: str,
-    peptide_sequence: str,
-    crosslinker_position_within_peptide: int,
-) -> int:
+def add_positions_of_amino_acid_where_crosslinker_bound_to_df(
```
jorisfu, line 108: Not quite sure about the semantics and use of this one. We modify `crosslinking_df`, but also return it. Hence pandas will override the passed df in-place but we also get a new reference, which we could (and do in this pr) use to override the old name.

So what I'm trying to say is we could also just only return messages here and not return the df, OR create a copy (probably not the preferred way here)

backend/protzilla/data_analysis/crosslinking_validation.py
```diff
@@ -78,86 +80,117 @@ def get_distance_between_two_amino_acids_in_angstrom(
...
-        )
-    )
-    amino_acid_position_crosslinker2_is_bound_to = (
-        get_position_of_amino_acid_crosslinker_bound_to(
-            protein_sequence=protein_sequence,
-            peptide_sequence=crosslink.Peptide2,
-            crosslinker_position_within_peptide=crosslink.CL_position2,
-        )
+    crosslinking_df["crosslinker_position1"] = pd.Series(
+        [pd.NA] * len(crosslinking_df), dtype="Int64"
```
jorisfu:
```suggestion
    crosslinking_df["crosslinker_position1"] = pd.Series(dtype="Int64")
```
Has the same effect and doesn't require list allocation (should probably be faster)

backend/protzilla/data_analysis/crosslinking_validation.py
```diff
@@ -78,86 +80,117 @@ def get_distance_between_two_amino_acids_in_angstrom(
...
     )
-    distance_in_alphafold = get_distance_between_two_amino_acids_in_angstrom(
-        amino_acid_position_crosslinker1_is_bound_to,
-        amino_acid_position_crosslinker2_is_bound_to,
-        protein_sequence[amino_acid_position_crosslinker1_is_bound_to - 1],
-        protein_sequence[amino_acid_position_crosslinker2_is_bound_to - 1],
-        cif_df,
+    crosslinking_df["crosslinker_position2"] = pd.Series(
+        [pd.NA] * len(crosslinking_df), dtype="Int64"
```
jorisfu: same here

backend/protzilla/data_analysis/crosslinking_validation.py
```diff
@@ -78,86 +80,117 @@ def get_distance_between_two_amino_acids_in_angstrom(
...
+            m.start() for m in re.finditer(f"(?={peptide_sequence2})", protein_sequence)
+        ]
+        all_position_combinations = [
+            (
+                pos1 + crosslinker_row.CL_position_within_peptide1 + 1,
+                pos2 + crosslinker_row.CL_position_within_peptide2 + 1,
+            )
+            for pos1 in peptide1_positions
+            for pos2 in peptide2_positions
+        ]
```
jorisfu, line 146: I understand correctly that we want a cartesian product of the two lists here, right? Python `itertools` provides a cleaner wrapper for that:


```suggestion
        peptide1_positions = [
            m.start() for m in re.finditer(f"(?={peptide_sequence1})", protein_sequence) +
            crosslinker_row.CL_position_within_peptide1 + 1
        ]
        peptide2_positions = [
            m.start() for m in re.finditer(f"(?={peptide_sequence2})", protein_sequence) +
            crosslinker_row.CL_position_within_peptide2 + 1
        ]
        all_position_combinations = list(itertools.product(peptide1_positions, peptide2_positions))
```

backend/protzilla/data_analysis/crosslinking_validation.py
```diff
@@ -78,86 +80,117 @@ def get_distance_between_two_amino_acids_in_angstrom(
...
+        if not all_position_combinations:
```
jorisfu:
```suggestion
        if all_position_combinations == [()]:
```
(if you opt to change it to itertools)

backend/protzilla/data_analysis/crosslinking_validation.py
```diff
@@ -78,86 +80,117 @@ def get_distance_between_two_amino_acids_in_angstrom(
...
+                msg = f"Peptide sequences {peptide_sequence1} and {peptide_sequence2} of crosslink entry {idx} were not found in the protein sequence. The entry was deleted."
+            else:
+                msg = f"Peptide sequence {peptide_sequence1 if not peptide1_positions else peptide_sequence2} of crosslink entry {idx} was not found in the protein sequence. The entry was deleted."
+            messages.append(dict(level=logging.WARNING, msg=msg))
+            rows_to_delete.append(idx)
+            continue
+        crosslinking_df.at[idx, "crosslinker_position1"] = all_position_combinations[0][
+            0
+        ]
+        crosslinking_df.loc[idx, "crosslinker_position2"] = all_position_combinations[
```
jorisfu: Why `.at` and then `.loc`?

backend/protzilla/data_analysis/crosslinking_validation.py
jorisfu: Agree with Anna, this is more efficient and a good solution. Adding a lot of rows in a loop like this directly causes a lot of memcpys because pandas reallocates the same df over and over again. Had this problem in Biomedical. once

backend/tests/protzilla/data_analysis/test_crosslinking_validation.py
```diff
@@ -113,3 +107,146 @@ def test_get_distance_between_two_amino_acids_in_angstrom():
...
+
+    protein_sequence = "XXXXXXXX"
+
+    df, messages = add_positions_of_amino_acid_where_crosslinker_bound_to_df(
+        df, protein_sequence
+    )
+
+    assert len(messages) == 1
+    assert messages[0]["level"] == logging.WARNING
+    assert "was not found" in messages[0]["msg"]
```
jorisfu: Not a fan of checking errors like this but it's the nature of the whole codebase, so it's fine xD

jorisfu (changes requested):
Sorry again for the late review, unfortunately I have some requested changes xD (mostly for the sake of code quality)

jorisfu (approved):
lgtm

PR #255, opened 2026-02-14
253 multimer prediction upload

### LINKED ISSUE #253: Multimer prediction upload
## Issue description
The user can upload multimer predictions from alphafold in a new step - persistently and only for one run. The user can also manage them in settings.

## Acceptance Criteria
- [ ] The user can upload multimer predictions in a new step and can choose whether to save them persistantly
- [ ] if the user saves them persistently, they are written on disk
- [ ] the uploads can be seen and deleted in settings
- [ ] the user can load data from disk into the run

Elena-kal:
## Description
fixes #253 
The user can now upload multimer upload predictions inside a run and in settings. The user can also delete these in settings just the same way the user interacts with monomer structures.
Similar to the monomer structures, the user can also load a multimer prediction from disk into a run.

## Changes
I renamed the protein structure upload of monomers into monomer structure upload across the codebase to be able to differentiate between monomer and multimer steps and processes. 
In [alphafold_protein_structure_load.py](backend/protzilla/importing/alphafold_protein_structure_load.py) I added the new functionality and refactored the existing code to make it reusable for the multimer upload. I added the new step in [importing.py](backend/protzilla/methods/importing.py). Here, I also added the step that lets the user load existing predictions from disk into the run.
For the management of multimer predictions in settings I added the file [multimer-structure-upload.tsx](frontend/src/components/app/settings/other-settings/multimer-structure-upload.tsx). I also refactored existing code and added new logic in [views_settings.py](backend/main/views_settings.py).

## Testing
To get the files you will need to upload, you need to first get a prediction form the alphafold server. For that go to [alphafoldserver.com](https://alphafoldserver.com) and log yourself in. You can make up to 30 prediction per day for free so no worries about running out of computing jobs. 
Here, you will need to upload the protein IDs of the proteins in the structure you would like a prediction of. I suggest you use hemoglobin alpha and hemoglobin beta (use two copies of each.) You will need to get their protein IDs and their sequences. In the end, you should have something like this:
 
Screenshot 2026-02-14 at 12 48 59 PM

Go on and click on continue and preview job. You will download a folder with several predictions. You will only need one though. You will need the cif file, confidence summary and full data summary. 
Furthermore, you will need to save the sequences as one fasta file that you can upload. 
Now you are good to go and start testing.

Upload a prediction inside a run - persistant and not. Check what happens if you upload something twice. Upload predictions from settings, see if they show up in the list. Maybe delete them.
Just play around and try breaking it, but not too maliciously :)) 
Lastly, do the same with the monomers. I have changed code here too, so it would be nice to check the functionality of these again. There are no new features for the monomers so simply test them like we have in the past.

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**


backend/main/views_settings.py
```diff
@@ -227,50 +233,123 @@ def save_ptm_settings(request, default_file_stem: str = DEFAULT_PTM_SETTINGS_FIL
...
+# <--- helper functions for monomer and multimer structure prediction --->
+def check_and_copy_files_to_directory(file_names: list, target_dir: str):
+    if target_dir.exists():
+        return False, "Entry ID is not unique."
+    else:
+        target_dir.mkdir(parents=True, exist_ok=True)
 
-AF_DICT_PATH = EXTERNAL_DATA_PATH / "alphafold"
+    for file_name in file_names:
+        source_dir = settings.FILE_UPLOAD_TEMP_DIR / file_name
```
jorisfu: `source_dir` is a file I suppose? Please rename to `source_file` then

backend/main/views_settings.py
```diff
@@ -227,50 +233,123 @@ def save_ptm_settings(request, default_file_stem: str = DEFAULT_PTM_SETTINGS_FIL
...
-    metadata_csv = AF_DICT_PATH / "alphafold_metadata.csv"
+def delete_structure(dir_path: str, csv_file_path: str, request):
+    if request.method != "POST":
+        return JsonResponse(
+            {"success": False, "message": "Invalid request method"}, status=405
+        )
+
+    data = json.loads(request.body)
+    entry_id = (str(data.get("entry_id") or "")).strip()
+    if not entry_id:
```
jorisfu, line 270: If so, we don't need the error handling in the previous line, we could just use `str(data.get("entry_id")).strip()`

backend/main/views_settings.py
```diff
@@ -279,38 +358,42 @@ def upload_prot_structure(request):
...
+            "gene",
+            "model_used",
+        ]
+
+        df = get_metadata_df(
+            csv_file_path=metadata_csv, expected_columns=expected_columns
+        )
 
         now_utc = datetime.now(timezone.utc)
         formatted = now_utc.strftime("%Y-%m-%dT%H:%M:%SZ")
```
jorisfu: Weird name for a variable, just do `timestamp = datetime.now(timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ")`

backend/main/views_settings.py
```diff
@@ -333,61 +416,111 @@ def upload_prot_structure(request):
...
+
+        df = get_metadata_df(
+            csv_file_path=metadata_csv, expected_columns=expected_columns
         )
 
-    # delete folder with files for the protein structure
-    target_dir = AF_DICT_PATH / entry_id.upper()
-    metadata_csv = AF_DICT_PATH / "alphafold_metadata.csv"
+        now_utc = datetime.now(timezone.utc)
+        formatted = now_utc.strftime("%Y-%m-%dT%H:%M:%SZ")
```
jorisfu: again, put this into one line

backend/main/views_settings.py
```diff
@@ -333,61 +416,111 @@ def upload_prot_structure(request):
...
+
+        new_row = {
+            "entry_id": entry_id,
```
jorisfu, line 512: uppercase here to avoid issues down the line (see main review comment)

backend/protzilla/importing/alphafold_protein_structure_load.py
```diff
@@ -330,96 +684,223 @@ def get_prot_structure_dfs(entry_id: str) -> dict[str, Any]:
...
+    messages = []
+
+    temp_dir, work_dir = get_correct_af_directories(
+        entry_id=entry_id,
+        directory_name=paths.ALPHAFOLD_MULTIMER_PATH,
+        persist_upload=persist_upload,
+    )
+
+    now_utc = datetime.now(timezone.utc)
+    formatted = now_utc.strftime("%Y-%m-%dT%H:%M:%SZ")
```
jorisfu: again date formatting

jorisfu (changes requested):
only minor code comments, see larger review comment

jorisfu:
Thoughts when testing from a user perspective:

## The Upload Form
- What exactly is the "Entry ID" and does it need to follow any specific format? I would prefer some tooltip or a default value to aid the user. Looks like it can just be any name we want to give the import
- How are the protein IDs delimited and what kind of protein ID (uniprot/gene name/something else) must be entered? The comma delimiting hint is only visible in the non-persistent upload
- Why and how do we need to enter the AlphaFold Model? Does that string need to conform to any format? Note that we **could** actually get this info from the AlphaFold CIF files themselves (e.g. citation metadata)
- How does the FASTA need to be formatted (turns out it needs the `>sp|` prefix for entries to work correctly). Do the provided protein IDs need to match to the fasta in the long run? What if they don't, and how would we validate that? Why don't we get the used protein IDs directly from the fasta if we need to upload that anyway and it provides IDs?

## Bugs
- The directories for the prediction files are in all-caps, but the managed entry_ids are not. Hence two entries point to the same directory. E.g. two uploads "Blub" and "blub" get registered separately but both use the directory "BLUB". I suggest to fix this by capitalising these IDs as soon as possible and only use uppercase (or lowercase) IDs
- The file upload dialogues only filter file types when using the persistent uploader, not the in-step uploader


Apart from that it seems to be working as expected.

Elena-kal:
> Thoughts when testing from a user perspective:
> 
> ## The Upload Form
> * What exactly is the "Entry ID" and does it need to follow any specific format? I would prefer some tooltip or a default value to aid the user. Looks like it can just be any name we want to give the import
> * How are the protein IDs delimited and what kind of protein ID (uniprot/gene name/something else) must be entered? The comma delimiting hint is only visible in the non-persistent upload
> * Why and how do we need to enter the AlphaFold Model? Does that string need to conform to any format? Note that we **could** actually get this info from the AlphaFold CIF files themselves (e.g. citation metadata)
> * How does the FASTA need to be formatted (turns out it needs the `>sp|` prefix for entries to work correctly). Do the provided protein IDs need to match to the fasta in the long run? What if they don't, and how would we validate that? Why don't we get the used protein IDs directly from the fasta if we need to upload that anyway and it provides IDs?
> 
> ## Bugs
> * The directories for the prediction files are in all-caps, but the managed entry_ids are not. Hence two entries point to the same directory. E.g. two uploads "Blub" and "blub" get registered separately but both use the directory "BLUB". I suggest to fix this by capitalising these IDs as soon as possible and only use uppercase (or lowercase) IDs
> * The file upload dialogues only filter file types when using the persistent uploader, not the in-step uploader
> 
> Apart from that it seems to be working as expected.

## The Upload Form
* You are right, the Entry ID does not have a form and can be anything the user wants it to be. We use it to ensure that the user can for example upload multiple predictions for the same protein. I do not see a reasonable default value as it totally depends on how the user would like to name the import. 
* The protein Ids are uniprot Ids but as we call them protein ids across the plattform I kept it consistent and named it protein id as well. I will add the comma delimiting hint to the persistant upload.
* The user needs to upload the alphafold model for the monomer uploads (chris requested that) and I considered it reasonable that we would like the same information for multimers. We do not need it anywhere specifically yet. Maybe we can talk about it tomorrow?
* The fasta needs the same formatting as anywhere else in out software. After talking to chris about it a few weeks ago, I assumed it reasonable to not write any hints as it is not done anywhere else where we import fasta files. I guess we can definitely validate the protein ids from the fasta with the ones entered specifically. I think that would be more intuitive than simply taking the protein ids from the fasta directly. the order is not important as it is turned into a dataframe. In most steps we just assume that the users input is correct and build upon that so I did not implement a validation. It is good to have though.
## BUGS: 
* Yes you are right, thanks!
* Tarek implemented this behavior for settings, I cannot use it for the in-step uploader though. I would argue if we want this, it should be another issue.

backend/main/views_settings.py
Elena-kal: I don't think so. What you propose could break if data.get("entry_id") = None. (str(data.get("entry_id") or "")).strip() would turn it into "". What you propose would turn None into "None". I do not think we want that behavior.

Elena-kal:
I actually decided to remove the .upper() completely. I do not see a concrete reason to keep it and I believe it would be more intuitive to make it case sensitive. What do you think? Should I add a hint that it is case sensitive? Or just the opposite, do you think always uppercase makes more sense and give a hint that the entry id will be saved as uppercase? I do not know actually...

NeleRiediger (approved):
Everything seems to work fine. While testing I just took some notes of little things that bothered me: 
- I actually agree with Joris, that it is not obvious that the Entry ID is not something specified, but can be chosen completely freely. Maybe there is a way to make that more obvious with a little note or something. 
- For the multimer upload in settings: maybe a hint in which format the protein ids should be entered for the multimer import (e.g. comma separated) (the same way it is already done in the corresponding steps)   
- If I import something with an incorrect fasta file in settings, that import is successful. But if I try to import that from disk into a run, I get an error for the incorrect fast file. Wouldn't it be cleaner if that error already occurred when trying to import in settings? 

The code itself looks fine too

jorisfu:
Please make sure we cannot create two imports that map to the same directory name (when uppercased) on the system. As an idea, create another metadata field for each persistent import to map to a directory name (always uppercased) and ensure that no new import maps to that directory. I remember some peculiar issues with Windows when having two directories with the same name but different cases

jorisfu (approved):

PR #256, opened 2026-02-15
233 density plot or histogram for crosslinking validation based on length deviation

### LINKED ISSUE #233: Density Plot or Histogram for crosslinking validation based on length deviation
## Issue description
Replace the simple bar plot with a density plot or histogram.

## Acceptance Criteria
- [ ] Display a density plot or histogram after the crosslinking validation with angstrom deviation run.
- [ ] Show Crosslinker length in plot.
- [ ] Show the bounds on the deviation in the plot.

AnnaPolensky:
## Description
fixes #233 
Added plots for the crosslinking validation based on length deviation. For each crosslinker, 2 histograms are generated: One contains all the crosslinks that match the predicted data and all the crosslinks that do not. The other histogram contains contains only those crosslinks with predicted length within the mean ± 2 standard deviations.
The crosslinker length is drawn as a vertical line and if the user specified bounds on the deviation these are drawn as a vertical line as well.
grafik

## Changes
The plot method diagrams_of_crosslinking_validation_data in backend/protzilla/data_analysis/crosslinking_validation.py was changed, all the calls to create_histogram happen there, so all the parameters are getting determined in this function.
Added more parameters to create_histogram in backend/protzilla/data_preprocessing/plots.py so that you can specify a range, whether each bin should have exactly the size of one int or what the relevant column is. Also made sure that the function won't crash if it receives an empty dataframe.

## Testing
Use the adapted crosslinking data and protein O43242 to see which diagrams you get when only setting some bounds, all bounds or changing the bounds to different values.

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**


backend/protzilla/data_analysis/crosslinking_validation.py
```diff
@@ -293,29 +297,41 @@ def check_crosslink(crosslink: pd.Series) -> pd.Series:
     return dict(crosslinking_result_df=checked_crosslinks_df, messages=messages)
 
 
-def bar_plot_of_valid_crosslinks(
+def diagrams_of_crosslinking_validation_data(
```
Elena-kal, line 300: Doesn't this function belong to plots.py? You can still input validated_df. cif_df and amino_acid_sequence_df are only used as input to get validated_df. It would seem cleaner to simply use validated_df as input and move this function entirely.

backend/protzilla/data_analysis/crosslinking_validation.py
```diff
@@ -324,20 +340,133 @@ def bar_plot_of_valid_crosslinks(
...
-                invalid_crosslinks,
-            ],
-            names_of_sectors=["Valid Cross-Links", "Invalid Cross-Links"],
-            heading="Cross-Links used for Validation",
-            y_title="Number of Cross-Links",
+        mean_predicted_lengths = crosslinker_df["alphafold_distance"].mean()
+        standard_deviation_predicted_lengths = crosslinker_df[
+            "alphafold_distance"
+        ].std()
+        mean_plus_minus_two_std_range = (
```
Elena-kal: I think it would be cleaner to simply have to variables that explicitely state whether it is mean_minus_two_std_range or mean_plus_two_std_range. That makes later usages more intuitive as well e.g. here:

`min_value=mean_plus_minus_two_std_range[0],
            max_value=mean_plus_minus_two_std_range[1],`

backend/protzilla/data_analysis/plots.py
```diff
@@ -466,3 +468,31 @@ def prot_quant_plot(
     )
 
     return dict(plots=[fig])
+
+
+def add_vertical_line_with_annotation_in_legend(
+    fig: Figure, dash: str, annotation: str, x_value: float, color: str = "blue"
```
Elena-kal: nitpick but could we use a slightly different color, maybe the blue we already use in other places as well. I feel like the last BP put a lot of effort to make Protzilla's frontend pretty and consistent in its color usage. I think I saw a colors defined somewhere, maybe use them?

backend/protzilla/data_preprocessing/plots.py
```diff
@@ -169,6 +171,11 @@ def create_histograms(
     x_title: str = "",
     visual_transformation: str = "linear",
     overlay: bool = False,
+    relevant_column_a: str = None,
+    relevant_column_b: str = None,
```
Elena-kal, line 175: Why do we use a and b? Is there a better way to name these?

backend/protzilla/data_preprocessing/plots.py
```diff
@@ -188,47 +195,69 @@ def create_histograms(
...
+    :param relevant_column_a: Which column of dataframe_a should be used for the histogram. If None, the default_intensity_column will be used.
+    :param relevant_column_b: Which column of dataframe_b should be used for the histogram. If None, the default_intensity_column will be used.
+    :param min_value: Where the first bin should start. If None, will be set to the minimum value of the two dataframes.
+    :param max_value: Where the last bin should end. If None, will be set to the maximum value of the two dataframes.
+    :param one_bin_per_int: If set to True, min_value will be rounded down to the next int and max_value will be rounded up to the next int and there will\
+    be max_value-min_value many bins.
+
+    :return: returns a histogram of the data
     """
     if visual_transformation not in {"linear", "log10"}:
```
Elena-kal, line 207: Don't we have Enums for "linear" and "log10"?

NeleRiediger (approved):
Looks good and seems to work as expected. 
Just a little request: I think it would be nice to make it a bit more obvious which is the corresponding crosslinker to each plot.

backend/protzilla/data_analysis/crosslinking_validation.py
```diff
@@ -324,20 +340,133 @@ def bar_plot_of_valid_crosslinks(
...
+                    dash="dash",
+                    annotation=f"allowed deviation lower bound",
+                    x_value=crosslinker_length - accepted_deviation_lower_bound,
+                )
+
+        figures.append(histogram_two_standard_deviations)
+        figures.append(histogram)
+
+    valid_crosslinks = (validated_df["Is_intra_crosslink"] == True).sum()
+    invalid_crosslinks = (validated_df["Is_intra_crosslink"] == False).sum()
```
Elena-kal: Why do we call this valid crosslinks but use validated_df and above we do this:
```
for crosslinker, crosslinker_df in validated_df.groupby("Crosslinker"):
        distances_valid = crosslinker_df.loc[
            crosslinker_df["valid_crosslink"] == True, "alphafold_distance"
        ]
        distances_invalid = crosslinker_df.loc[
            crosslinker_df["valid_crosslink"] == False, "alphafold_distance"
        ]
        df_valid = pd.DataFrame({"alphafold_distance": distances_valid})
        df_invalid = pd.DataFrame({"alphafold_distance": distances_invalid})
```
Here we use crosslinker_df["valid_crosslink"] and not validated_df["Is_intra_crosslink"]. Are these the same? Why the same naming for different things? Why use different things for bar plot vs histogram?

backend/protzilla/data_analysis/crosslinking_validation.py
```diff
@@ -324,20 +340,133 @@ def bar_plot_of_valid_crosslinks(
...
+        mean_plus_minus_two_std_range = (
+            max(0, mean_predicted_lengths - 2 * standard_deviation_predicted_lengths),
+            mean_predicted_lengths + 2 * standard_deviation_predicted_lengths,
         )
-    ]
+        histogram_two_standard_deviations = create_histograms(
+            dataframe_a=df_valid,
+            dataframe_b=df_invalid,
+            name_a="Valid Crosslinks",
+            name_b="Invalid Crosslinks",
```
Elena-kal, line 399: That is super confusing, why name it a and b when they are valid and invalid crosslinks?

backend/protzilla/data_analysis/crosslinking_validation.py
```diff
@@ -324,20 +340,133 @@ def bar_plot_of_valid_crosslinks(
...
+            heading=f"Predicted distances for {protein_to_validate}, mean +- 2 standard deviations",
```
Elena-kal:
```suggestion
            heading=f"Predicted distances for {protein_to_validate}, mean +/- 2 standard deviations",
```

backend/protzilla/data_analysis/crosslinking_validation.py
```diff
@@ -324,20 +340,133 @@ def bar_plot_of_valid_crosslinks(
...
-    return [
-        create_bar_plot(
-            values_of_sectors=[
-                valid_crosslinks,
-                invalid_crosslinks,
-            ],
-            names_of_sectors=["Valid Cross-Links", "Invalid Cross-Links"],
-            heading="Cross-Links used for Validation",
-            y_title="Number of Cross-Links",
+        mean_predicted_lengths = crosslinker_df["alphafold_distance"].mean()
```
Elena-kal: Why plural? its just one value right?

backend/tests/protzilla/data_analysis/test_crosslinking_validation.py
```diff
@@ -247,3 +251,270 @@ def test_add_crosslinker_positions_with_overlapping_peptide_matches():
...
+
+
+@pytest.fixture
+def sample_crosslinking_df_with_no_std():
+    return pd.DataFrame(
+        {
+            "Crosslinker": ["CL1", "CL1", "CL2", "CL2"],
+            "alphafold_distance": [10.5, 10.5, 10.5, 10.5],
+            "valid_crosslink": [True, False, True, False],
+            "Is_intra_crosslink": [True, False, True, False],
```
Elena-kal: Just curious: is valid_crosslink and Is_intra_crosslink always the same? What would happen if they were not?

Elena-kal (changes requested):
I left some comments on the code. 
I would also suggest different headdings for the different plots. Right now they are the same for the different crosslinks. Maybe put the crosslinker name in the headding as well. I found it a bit unintuitive to only see the difference between the plots in the legend.

Also with the valid and invalid crosslinks you can click on the names in the legend and they disappear in the plot so the user can see the plot without them. The vertical lines are also clickable in the legend but nothing happens. That seems unintuitive as well. Would be nice to be able to remove them as well.

backend/protzilla/data_analysis/crosslinking_validation.py
AnnaPolensky: crosslinker_df is a subset of validated_df for a specific crosslinker. But very fair point, it should be valid_crosslink as well.

backend/protzilla/data_preprocessing/plots.py
AnnaPolensky: Because they belong to dataframe_a and dataframe_b and the create_histogram-function is not specific to crosslinking.

backend/tests/protzilla/data_analysis/test_crosslinking_validation.py
AnnaPolensky: No, is_intra_crosslink only describes whether we have a crosslink between two different proteins or not, valid_crosslink describes whether the crosslink matches the prediction. For these tests, we actually only need valid_crosslink, so I removed is_intra_crosslink.

PR #258, opened 2026-02-15
use step outputs for protein_df

jorisfu:
## Description
fixes #257 
Changes `get_step_input` to `get_step_output` for the relevant steps

## Testing
Use the AML Paper workflow found [here](https://nextcloud.hpi.de/apps/files/files?dir=/PROTzilla_2025). Add an imputation step to the end of the data preprocessing section. Perform that, then perform the t-test. Note how there are no missing samples / NaN values left compared to the previous state

## PR checklist
**Development**
- [x] If necessary, I have updated the documentation (README, docstrings, etc.)
- [x] If necessary, I have created / updated tests.
 
**Mergeability**
- [x] main-branch has been merged into local branch to resolve conflicts
- [x] The tests and linter have passed AFTER local merge
- [x] The backend code has been formatted with `black`
 
**Code review**
- [x] I have self-reviewed my code.
- [ ] At least one other developer reviewed and approved the changes

PR #260, opened 2026-02-15
259 add imputation to hierarchical clustering

### LINKED ISSUE #259: Add Imputation to Hierarchical Clustering
## Issue description
The Plot: Clustergram does not support NaN values in dataframes. In the current list based workflow, there is no way to use imputation only for the clustergram. Since an imputation step severely skews the t-test, this is the only clean-ish way to get a clustergram matching the AML paper a bit more closely.

## Acceptance Criteria
- [ ] When clustering data that was previously imputed, everything works as previously.
- [ ] When clustering data that contains NaNs, you get a warning but still a clustergram. You can also change the imputation mode in the form and get quite different results lol

jorisfu:
## Description
fixes #259 
Adds an imputation strategy input to the clustergram plot.

## Changes
Mostly in `data_analysis.py`, adjusting the form. Note that this also includes the quick fix in #258 

## Testing
Use the AML paper workflow from [here](https://nextcloud.hpi.de/apps/files/files?dir=/PROTzilla_2025) and check the acceptance criteria in https://github.com/cschlaffner/PROTzilla/issues/259

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**


backend/protzilla/data_analysis/plots.py
```diff
@@ -277,6 +289,10 @@ def clustergram_plot(
         else:
             custom_color_scale = None
 
+        imputer_parameters = dict(
+            axis=AXIS_PROTEIN, missing_values="nan", strategy=imputation_strategy
```
Elena-kal, line 301: Just wondering: Is it alway "nan"? Could it ever be "NaN" or something like this?

Elena-kal (approved):
Code looks good, also worked when tested. I only have one small issue with the frontend part of it  (maybe we could consider making an issue of this if this seems relevant). When the user runs the clustergram on a slightly smaller screen (e.g. my laptop screen 13,6-inch) there is no way to look at the entire plot. It is cut off and one cannot scroll horizontally.  It looks like this below.

Image

The only way to look at the plot is to make the window larger than the screen and cutting it off on the other side like this:

Image

It's not the end of the world but it is slightly annoying.

hendraet:
> Code looks good, also worked when tested. I only have one small issue with the frontend part of it (maybe we could consider making an issue of this if this seems relevant). When the user runs the clustergram on a slightly smaller screen (e.g. my laptop screen 13,6-inch) there is no way to look at the entire plot. It is cut off and one cannot scroll horizontally. 

Yeah, that is an issue with multiple plots in PROTzilla and should probably become a separate issue. This also affects the PTM visualizations for example. (As a workaround, it's sometimes enough to hide the navigation on the left)

backend/protzilla/data_analysis/plots.py
jorisfu: it's actually always `np.nan`, but the Clustergram class needs this kind of string as an input to validate the missing_values parameter. Felt cleaner to just pass it here than adjusting the Clustergram class

backend/tests/protzilla/data_analysis/test_plots_data_analysis.py
```diff
@@ -182,9 +182,8 @@ def test_clustergram_nans_in_input(wide_4d_df):
 
     outputs = clustergram_plot(nan_df, metadata_df=None, flip_axes=False)
     assert "messages" in outputs
-    assert "plots" not in outputs
     assert any(
-        "The selected input dataframe contains missing values." in message["msg"]
+        "Input contains missing data" in message["msg"]
```
NeleRiediger: Just a really minor issue, but the wording seems somewhat strange, it's a contradiction in itself. Not sure why you changed it, but the old wording of "contains missing values" sounds a little clearer. If the new wording was supposed to convey some more meaning, I didn't understand it, which is probably also a reason to make it a bit more explicit. But again, this is really just nitpicking.

NeleRiediger (approved):
Testing delivered the expected results and the code seems fine too.

backend/tests/protzilla/data_analysis/test_plots_data_analysis.py
jorisfu: adjusted

Issue #250, opened 2026-02-10
Backend Graph Representation


jorisfu:
## Issue description
The Graph representation needs to be explicitly modelled in the back-end.

## Acceptance Criteria
- [ ] There is a graph representation of the connected steps
- [ ] Based on this graph representation, the invalidation of steps gets updated
- [ ] There is a toposort representation for the list editor (frontend) and the next button

jorisfu:
Closing as we do this in #179

Issue #179, opened 2026-01-08
Refactor `Step` to allow selection of step to retrieve dataframe from


tE3m:
## Issue description

Currently, most insert the "global" dataframes that the stepmanager provides. In order to allow selecting the source of input dataframes in the frontend, each steps needs to be able to store the locations that its input(s) can be found at

## Acceptance Criteria
- [ ] Valid dataframe keys defined in constants
- [ ] API exposes ability to set which step another step's input is taken from
- [ ] all steps insert their input dfs from the specified steps
- [ ] documentation of valid dataframe schemes

PR #262, opened 2026-02-15
261 node status updates one click late

### LINKED ISSUE #261: Node status updates one click late
## Issue description
Node status in the Graph updates with a one‑click delay when navigating between nodes. The status (e.g., pending/outdated) only becomes visible after an additional click or interaction.

### Expected Behavior
Status color updates immediately when a node is selected.

## Acceptance Criteria
- [x] Status color updates on node selection.
- [x] No extra click or interaction is required for the status update.

3dot141592:
## Description
fixes #261
- Prevented out-of-order `getRunData()` responses from only allowing newer state to overwrite

## Changes
- Guarded `getRunData()` with a request ID to ignore stale responses.

## Testing
1. Open Node Editor.
2. Select a node.
3. Status color updates immediately without a second click.

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**


frontend/src/components/app/run-screen/node-editor/node-editor.tsx
```diff
@@ -316,14 +316,14 @@ export const NodeEditor: React.FC<NodeEditorProps> = ({
...
-            1
+              1
           }
           onNext={() => {
             console.log("TODO: A vulture ate this callback! Come up with something better.");
           }}
           onSubmit={onFormSubmit}
           onChange={() => {
-            console.log("TODO: A vulture ate this callback! Come up with something better.");
+            navigateOrRefreshSteps();
```
jorisfu, line 378: As far as my testing goes, this line alone fixes the entire problem. The outdated_count should not make any difference here

jorisfu (changes requested):

frontend/src/components/app/run-screen/node-editor/node-editor.tsx
jorisfu: Tbf we need a more robust solution for reloading the state to the front-end anyway, but that can wait for some other PR after #250 and #179 are complete

AnnaPolensky: So should we remove outdated_count or might it be useful for something else?

AnnaPolensky (approved):
Could not test locally because some file paths contain ":" which is a problem for Windows (could not checkout the branch).

frontend/src/components/app/run-screen/node-editor/node-editor.tsx
3dot141592: No, as far as i understand the problem, this line does not work on its own. Still, the way i implemented it wasnt good either. I have looked into this further and the problem are race conditions: an old call overwrites the new one, resulting in an outdated form. I have fixed this in my newest push and removed the old outdated_count stuff.

jorisfu (approved):
lgtm

frontend/src/components/app/run-screen/run-screen.tsx
```diff
@@ -115,11 +116,14 @@ export const RunScreen: React.FC = () => {
   };
 
   const getRunData = useCallback(async () => {
+    const requestId = (runDataRequestID.current += 1);
     const response = await callApiWithParameters("get_run_data/", {
       run_name: runName,
     });
     if (response) {
-      setRunData(response.data);
+      if (requestId === runDataRequestID.current) {
```
AnnaPolensky, line 125: nit: couldn't we combine the two ifs? `if (response && requestId === runDataRequestID.current)`

AnnaPolensky (approved):
Looks good

jorisfu:
There were some general adjustments necessary in the node editor (see #272 and [this commit](https://github.com/cschlaffner/PROTzilla/commit/39713d355bd81c5aa037f79814fbabb154071b33)]) that should've fixed this as a side effect. We should keep this approach in mind for the future but I would not merge this simply because the problem seems to not occur currently

PR #268, opened 2026-02-26
feat: add first draft for visualization

### LINKED ISSUE #247: Add visualization for protein structures
## Issue description
Whenever we work with protein structure data, a visualization of the structures should be provided.  

## Acceptance Criteria
- [ ] When importing a protein structure prediction users should be able to visually see the structure they just imported
- [ ] When validating this imported structure prediction they should see this as well extended to include the crosslinks used for validation

NeleRiediger:
## Description
fixes #247 
I created a new frontend-component for the Mol*-Viewer. This component is used as a possible output of a step (just as plots or tables) on the run-screen where it gets visualization-data (in the form of a cif converted to a string from its dataframe representation) whenever a step has visualization-data as one of its output keys (everything adapted to the newly refactored output system). 
All steps that import a protein structure or validate crosslinks have visualization-data as one of their output-keys now. For the validation steps the visualization data not only consists of the cif-dataframe but also the crosslink-dataframe. 
Whenever the Mol*-Viewer component also gets crosslinking-data the component creates a cif-structure for those crosslinks and adds it to the viewer as a separate structure on top. 
Since dealing with all this cif data and loading the viewer can sometimes take some time, I then added some functionality to the run-screen, that enables visualizations to only get loaded when the user actually clicks on the visualization output tab and never in the background. So the user doesn't experience significant delays in the run because of it. (Visualizations that are loaded, are also cached, while the step doesn't change).  
Visualization should work for both monomers and multimers. But inter-crosslinks in multimers need some additional information to be displayed correctly. Since this information is also necessary for a correct validation of multimer crosslinks and this functionality needs to be implemented there anyways, I figured we could just calculate it there and output it with the visualization data. So as of now I just hardcoded this information and provided the interface for when it is implemented in the future. That's why inter-crosslinks in multimeres are wrongfully displayed as intra crosslinks right now. 

## Changes
Mol*-Viewer frontend component: 
- molstar-viewer.tsx ➞ base-code to initialize the viewer and display a structure in it 
- molstar-viewer.service.tsx ➞ some helper functions for the viewer 
- crosslinker-processing ➞ the functions creating a cif-structure for the crosslinks 
- (and some more like index, props, config, styles)    
 
Where I added visualization-data as an output key: 
- alphafold-protein-structure-load.py 
- cross linking-validation.py 

Getting the visualization-data:
- run-screen.tsx ➞ added visualizations as an output-type (within the newly refactored output system) and added the lazy-loading for visualizations (including a selectedOutputTab-state) 
- views.py and views-helper.py ➞ functionality that extracts the right kind of data from the step-output, when a visualization is created  

## Testing
- create a standard-crosslinking workflow (up to validation) for both monomers and multimers and check wether all steps that load or validate structures have a visualization and that it looks correct 
- when validating crosslinks use some different values so the same crosslinks are once rated valid and once invalid, then you should see them visualized in a stronger or more pale color 
- (keep in mind that inter-crosslinks are displayed as intra-crosslinks even though they already have the purple color indicating they are inter) 
- maybe play around with switching between different output tabs to make sure the selectedOutputTab-state is working as expected  

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**


jorisfu:
I just skimmed over the PR and I think I need a bit more context as to what we're trying to do here. Adding another attribute to the step surely works, but I think it's not very clean.

As a suggestion, why don't we refactor the current output system to specify the types of the outputs and then just use that? We could e.g. add another specific output type for this kind of visualisation and then add another tab in the frontend outputs dynamically if such an output exists for a given step.

Just a general idea though. Not entirely sure how to concretely implement it yet due to lack of context, but let's discuss this in a meeting

backend/main/views.py
```diff
@@ -3,6 +3,8 @@
 import traceback
 from zipfile import ZipFile
 import re
+import traceback
+from typing import Any
```
jorisfu: already on line 9

frontend/src/components/app/run-screen/run-screen.tsx
```diff
@@ -341,6 +382,20 @@ export const RunScreen: React.FC = () => {
...
+  const visualizationComponent = (
+    <StyledContentContainer>
+      {visualizations.length > 0 ? (
+        visualizations.map((viz) => (
+          <StyledContentDiv key={viz.proteinEntryId}>
+            <MolstarViewer cifText={viz.cifString} crosslinks={viz.crosslinks} />
+          </StyledContentDiv>
+        ))
+      ) : (
+        <SectionTitle baseComponent="h4" description="Structure-visualization is loading..." />
```
jorisfu: "Structure visualisation is loading" fix for EN-UK

frontend/src/components/app/run-screen/run-screen.tsx
```diff
@@ -359,10 +414,7 @@ export const RunScreen: React.FC = () => {
           }))}
         />
       ) : (
-        <SectionTitle
-          baseComponent={"h4"}
-          description={"This step does not provide any tables as output."}
-        />
+        <SectionTitle baseComponent={"h4"} description={"Output-tables are loading..."} />
```
jorisfu: "Output tables are loading" no hyphen here

frontend/src/components/app/run-screen/run-screen.tsx
```diff
@@ -437,8 +489,20 @@ export const RunScreen: React.FC = () => {
     availableTables && availableTables.length > 0 && { name: "Tables", value: tableComponent },
     availableImages.length > 0 && { name: "Images", value: imageComponent },
     availableDownloads.length > 0 && { name: "Downloads", value: downloadComponent },
+    availableVisualizations.length > 0 && { name: "Visualizations", value: visualizationComponent },
```
jorisfu: "Visualisations" for EN-UK

frontend/src/components/app/run-screen/run-screen.tsx
```diff
@@ -471,6 +535,14 @@ export const RunScreen: React.FC = () => {
                 styleProps={{ height: "calc(100% - 3em)" }}
                 components={components}
                 hasCardTitle={false}
+                selection={selectedOutputTab}
+                callback={(component) => {
+                  setSelectedOutputTab(component.name);
+
+                  if (component.name === "Visualizations" && !hasLoadedVisualizations) {
```
jorisfu: "Visualisations"

frontend/src/components/core/shared/molstar-viewer/molstar-viewer.tsx
```diff
@@ -0,0 +1,81 @@
...
+          handleError(disposeError, "Error disposing Molstar plugin:", notify);
+        }
+      }
+    };
+  }, [cifText, crosslinks, notify]);
+
+  return (
+    <Container>
+      {isLoading && (
+        <SectionTitle baseComponent="h4" description="Structure-visualization is loading..." />
```
jorisfu: "Structure visualisation is loading"

jorisfu (changes requested):
Tested working and seems all good apart from some external facing messages needing to be adjusted.

The only major problem left is the lack of proper serialisation/deserialisation in the disk operator leading to visualisations bloating the `run.yaml`, causing horrible performance due to repeated re-writes. We should only update visualisation dumps if the output changed and we thus must also dump them into another file. See e.g. the changes I did with PNGs on this PR for a reference: https://github.com/cschlaffner/PROTzilla/pull/302

backend/main/views.py
```diff
@@ -715,6 +717,37 @@ def get_downloads_from_step(request: HttpRequest):
...
+        output_key = data.get("output_key")
+
+        run = Run(run_name)
+        step = run.steps.get_step_by_id(step_id)
+        visualization_dict = step.output.get(output_key)
+        if visualization_dict is None:
+            return JsonResponse(
+                {
+                    "success": False,
+                    "message": "Got no available visualization for the step",
```
Elena-kal, line 734:
```suggestion
                    "message": "No visualization available for this step.",
```

backend/main/views.py
```diff
@@ -715,6 +717,37 @@ def get_downloads_from_step(request: HttpRequest):
...
+                    "data": {},
+                }
+            )
+        return JsonResponse(
+            {
+                "success": True,
+                "message": "Got the available visualization for the step",
```
Elena-kal, line 741:
```suggestion
                "message": "Visualization loaded for this step.",
```
"got" sounds a bit informal

backend/main/views_helper.py
```diff
@@ -176,3 +178,113 @@ def load_yaml_from_file(path: Path) -> str:
...
     with path.open("r") as f:
         return f.read()
+
+
+# ------------------------- helper for get_step_visualization: -------------------------
+
+
+def create_visualization(
+    cif_df: pd.DataFrame,
+    protein_entry_id: str,
```
Elena-kal: In previous code we often called this protein_id or uniprot_id, is there a reason why it is now protein_entry_id? Does it have a different function?

backend/main/views_helper.py
```diff
@@ -176,3 +178,113 @@ def load_yaml_from_file(path: Path) -> str:
...
+    crosslinking_df: Optional[pd.DataFrame] = None,
+) -> dict:
+    """
+    Convert a CIF DataFrame to a mmCIF string and package it with its protein entry ID.
+    Optionally include crosslinks.
```
Elena-kal, line 193: Feels weird to call the function something with visualization but not to mention the visualization part in the doc string.

backend/main/views_helper.py
```diff
@@ -176,3 +178,113 @@ def load_yaml_from_file(path: Path) -> str:
...
+
+    result = {"proteinEntryId": protein_entry_id, "cifString": cif_string}
+
+    if crosslinking_df is not None:
+        result["crosslinks"] = extract_relevant_crosslink_information(crosslinking_df)
+
+    return result
+
+
+def convert_df_to_mmcif_for_visualization(
```
Elena-kal:
```suggestion
def convert_cif_df_to_mmcif_for_visualization(
```

backend/main/views_helper.py
```diff
@@ -176,3 +178,113 @@ def load_yaml_from_file(path: Path) -> str:
...
+
+    cif_string = "\n".join(lines)
+    return cif_string
+
+
+def extract_relevant_crosslink_information(
+    crosslinking_df: pd.DataFrame,
+) -> List[Dict[str, int]]:
+    """
+    For each crosslink extract its relevant information from a DataFrame.
```
Elena-kal, line 263: It is fine not to mention what the relevant information is in the name of the function. I think it would be helpful though to have it in the doc string.

jorisfu (changes requested):
Some necessary general changes:

- Just use the ArtifactOperator for dumping that @Elena-kal wrote to avoid repeating code, it was intended for things like this
- Since your visualisations are generated by a calculation method, they are regular outputs. The `self.artifact_versions["visualization"]` does not make any sense like this because it never gets updated in your code anyway as the visualisation version counts are updated with `self.artifact_versions["output"]` (see `steps.py:155`). You can safely remove all code related to that key.

So TL;DR you only need the two case statements in `_read_outputs` and `_write_output` and you can reuse everything else as-is.

Will test the adjusted version then but I suppose this should generally work

backend/protzilla/disk_operator.py
```diff
@@ -478,10 +486,26 @@ def _write_output(self, step: Step) -> dict:
...
+                        output_data[key] = OutputItem(
+                            output_type=OutputType.VISUALIZATION,
+                            value=str(file_path.relative_to(self.run_dir)),
+                        )
                     case _:
                         output_data[key] = item
 
             self._update_dump_state(step, "output")
+            # self._update_visualization_dump_state(step)
+            if visualization_written:
```
Elena-kal: I am a bit confused on when we set `visualization_written` to true. As far as I see it, it can either be false or it is never set, right?

frontend/src/components/core/shared/molstar-viewer/crosslinker-processing.tsx
```diff
@@ -0,0 +1,245 @@
...
+): CrosslinkerAtom | null {
+  const lines = cifString.split(/\r?\n/);
+
+  for (const line of lines) {
+    const trimmed = line.trim();
+    if (!trimmed || trimmed.startsWith("#")) continue;
+    const tokens = trimmed.split(/\s+/);
+    if (tokens[0] !== "ATOM" && tokens[0] !== "HETATM") continue;
+
+    const cifIndices = getCifAtomSiteIndices(cifString);
```
Elena-kal: Do we want to calculate these indices for each line? Wouldn't it be enough to do it once?

frontend/src/components/core/shared/molstar-viewer/crosslinker-processing.tsx
```diff
@@ -0,0 +1,245 @@
...
+  };
+  return mapping[reactiveAtom] || "CA";
+}
+
+function findAtomCoordinatesInCif(
+  cifString: string,
+  crosslinkerAtomId: string,
+  crosslinkerSeqPos: number,
+  crosslinkerChainId: string,
+): CrosslinkerAtom | null {
```
Elena-kal: I think the `| null` is a bit misleading. It either returns an atom or it throws an error, it will never return null, will it?

frontend/src/components/core/shared/molstar-viewer/crosslinker-processing.tsx
```diff
@@ -0,0 +1,245 @@
...
+
+      const connectionLine = [
+        `${String(connectionId)} misc ${atom1Id}`,
+        `X CL ${String(atom1.seqPos)} ${atom2Id}`,
+        `X CL ${String(atom2.seqPos)}`,
+      ].join(" ");
+
+      connectionLines.push(connectionLine);
+
+      console.log(connectionLine);
```
Elena-kal: Do we need this log?

Elena-kal (commented):
Overall, the code looks clean and good. I added a few remarks here and there. Tested it and it worked fine. Good work!

Also, a few of my comments are really nitpicky and do not need to be addressed, they are just what I noticed but we do not need to be perfectionists all the time :))

backend/main/views.py
NeleRiediger: First of all, I don't have a strong opinion about these messages, it's absolutely fine with me to change them. I just want to briefly point out that I chose the wording to be consistent with the existing messages for plots, tables, and other outputs. So, if we change it, we should probably do the same for every other output for the sake of consistency. It's not a huge amount of work, but what do you think?

backend/main/views.py
NeleRiediger: Same as above

backend/main/views.py
jorisfu: Just FYI these messages are never displayed anywhere, you could write literally anything into these. What @NeleRiediger wrote is consistent with the rest of the codebase so it's ok IMHO.
If anything, we should remove these success messages consistently and rely on error codes + error messages instead. Same goes with the "success" variable (not like we have return codes that indicate success anyway???)

Sorry for rant lol

backend/main/views_helper.py
NeleRiediger: My initial thought was that you used an entry id for multimeres as a truly unique identification for any uploaded structure. So I thought using this for multimeres instead of their uniprot ids would be a good idea and I thought the variable name entry id would convey that it is supposed to be a unique designation für the structure. 
However, since using the actual entry id as an output of the validation function would have required major code changes, I ended up using a list of the uniprot ids anyways. So I think we could just rename the variable. 
(Although I just noticed that for the upload functions I indeed used the actual entry ids of multimeres. So I guess we should rather talk about whether we want to use the uniprot ids or entry ids and then make everythink (including the variable name) consistent.) Do you have an opinion on that?

backend/main/views_helper.py
NeleRiediger: Makes sense, changed it

backend/main/views_helper.py
NeleRiediger: also changed that

backend/main/views_helper.py
NeleRiediger: Agree, added that

frontend/src/components/core/shared/molstar-viewer/crosslinker-processing.tsx
NeleRiediger: Totally agree, corrected that

backend/protzilla/disk_operator.py
NeleRiediger: Might have forgotten to remove this code. I was trying something, but decided against it in the end

frontend/src/components/core/shared/molstar-viewer/crosslinker-processing.tsx
NeleRiediger: Agreed

frontend/src/components/core/shared/molstar-viewer/crosslinker-processing.tsx
NeleRiediger: No, forgot to remove that

backend/main/views.py
Elena-kal: I do not have a really strong opinion about them either, especially if the user never sees them. Then it is actually better how you did it to keep the wording consistent with the rest of the code. Personnally, I do not like the word "get" thats all, but it really does not matter. Lets just drop this comment here and see it as resolved.

backend/main/views_helper.py
Elena-kal: Well, the reason because we have entry ids is that we can have multiple predictions for the same structure. I believe we have entry ids for monomers as well because they can also have multiple predictions. That is why I would definetly use the entry ids. 
I think then I would say that the wording for the variable is fine. I believe it just irritated me because I thought it meant the protein id but this makes much more sense. Only for multimer structures where we have multiple proteins the name could be a bit misleading I guess.

backend/protzilla/disk_operator.py
```diff
@@ -441,6 +447,8 @@ def _write_output(self, step: Step) -> dict:
         """
         with ErrorHandler(), step.disk_write_mutex:
             output_data: dict[str, OutputItem] = {}
+            visualization_written = False
+            logger.info(f"Step outputs: {[k for k, v in step.output]}")
```
jorisfu: remove these two lines, apart from that lgtm and ok to merge

jorisfu (changes requested):

PR #269, opened 2026-03-01
v1.0.0 Release Branch

### LINKED ISSUE #179: Refactor `Step` to allow selection of step to retrieve dataframe from

jorisfu:
## Description
Large parts of the existing step management were overridden to ensure a more scalable software package that allows for steps to be loosely coupled and connected in a semantically meaningful fashion. This PR also includes the new Graph-based editor.

## Changes
- Step management was rewritten from a list-based to a MultiDiGraph-based approach
- Steps receive their inputs via connected steps and form inputs only, no implicit inputs anymore
- Node Editor allows workflow edits

## Caveats
- The runner is still broken and needs to be adjusted
- The list editor is broken and needs to be adjusted
- We still need some more test cases

## Testing
Create a new run with the standard worflow. Test around whether or not you can replicate all prior functionality with the new editor. Also please try to break it as badly as you can!

## PR checklist
**Development**
 
**Mergeability**
- [x] dev-branch has been merged into local branch to resolve conflicts
 
**Code review**


AnnaPolensky:
Looks very very nice. Some notes though:
- Sometimes nodes vanish. (Sometimes first the edges vanish and then with the next click the nodes vanish as well. I was not able to reproduce this behaviour consistently. It sometimes occured when spam-clicking a node, sometimes when switching back and forth between nodes you cannot calculate because they are dependent on not calculated nodes, but clicking calculate everytime you switch)
- Do we want to zoom in to a node everytime we click the node 3+ times?
- Consistent/meaningful determination of the next step when one clicks next. (e.g. not going to a step if not all dependencies are met if there is a step where all dependencies are met)
- Volcano Plot is not connected to anything in the standard workflow.
- Adding a Peptide-Import Step and clicking on it does not open the form for the peptide-Import.
- How to deal with 2 possible inputs, when we only will use one at a time. -> e.g. Filter Proteins: By samples missing
- Could we rename the "node editor fallback" to "node editor loading screen"?

Elena-kal (commented):
Overall I really like the nodes editor, although I think it still needs some improvements. Apart from the caveats you already noted and the ones Anna mentioned, I have the following observations:

- Do we always need a success message when a run is started successfully? It feels redundant and could probably be reduced or shown only when something meaningful changes.
- When trying to delete the first step, the UI shows “successfully deleted”, but the node is not removed in the node tab. If I try to delete it again, I get an error message. However, once I switch to another step, the node disappears. 
Image

- When switching to the volcano plot step, the tab does not load. It looks like there might be a problem in the form handling or form initialization for that step.
- Is the current appearance of the Loading tab intentional? If yes, it would be good to improve the visual presentation to make it clearer and more consistent with the rest of the UI.
Image

- While clicking through the interface, I encountered the following backend error:
`[2026/03/03 - 16:23:51] [ ERROR] Internal Server Error: /api/get_step_form/
File "/usr/local/lib/python3.11/site-packages/django/core/handlers/exception.py", line 55, in inner
response = get_response(request)
File "/usr/local/lib/python3.11/site-packages/django/core/handlers/base.py", line 197, in _get_response
response = wrapped_callback(request, *callback_args, **callback_kwargs)
File "/home/prot/zilla/backend/main/views.py", line 649, in get_step_form
form = run.current_form(new_form_values)
File "/home/prot/zilla/backend/protzilla/run.py", line 158, in wrapper
result = func(self, *args, **kwargs)
File "/home/prot/zilla/backend/protzilla/run.py", line 343, in current_form
self.steps.current_step.modify_form(self)
File "/home/prot/zilla/backend/protzilla/methods/data_analysis.py", line 769, in modify_form
run.steps.get_instance_identifiers(
AttributeError: 'StepManager' object has no attribute 'get_instance_identifiers'`

- How can Imputation produce a peptide dataframe as output without having a peptide dataframe as input? The same applies to Normalization. The input and output definitions do not seem consistent.
- I was not able to test much further because it is currently not possible to add new steps beyond those in the standard workflow. As a result, the standard workflow itself cannot be tested thoroughly.
- I also noticed that nodes and connections occasionally disappear from the graph.

jorisfu:
@Elena-kal 

> When trying to delete the first step, the UI shows “successfully deleted”, but the node is not removed in the node tab. If I try to delete it again, I get an error message. However, once I switch to another step, the node disappears.

Fixed that.

> Do we always need a success message when a run is started successfully? It feels redundant and could probably be reduced or shown only when something meaningful changes.

Not something we introduced, but agree.

> When switching to the volcano plot step, the tab does not load. It looks like there might be a problem in the form handling or form initialization for that step.

Yes, @tE3m 

> Is the current appearance of the Loading tab intentional? If yes, it would be good to improve the visual presentation to make it clearer and more consistent with the rest of the UI.

It's literally just a fallback, we might want to come up with something better (proper loading screen) indeed. Maybe another day...

> How can Imputation produce a peptide dataframe as output without having a peptide dataframe as input? The same applies to Normalization. The input and output definitions do not seem consistent.

@tE3m 

> I also noticed that nodes and connections occasionally disappear from the graph.

We should've hopefully fixed this (or, well, let Codex fix it because we have absolutely no idea how this stuff works. It's reactflow internals just messing with us). If there are any **specific** nodes/edges disappearing, please tell us how to reproduce that.

backend/protzilla/step_manager.py
```diff
@@ -0,0 +1,510 @@
...
+        self.disk_operator: DiskOperator = disk_operator
+
+        # Saves all steps, accessible by their instance identifiers
+        self.all_steps: dict[StepID, Step] = {}
+
+        # Graph saving connections between steps.
+        # All steps are represented by their instance_identifiers as nodes.
+        # The edges specify which source/target handles (or outputs/inputs)
+        # they refer to as a dict {"source_handle": <>, "target_handle": <>}
+        # in their data attribute
```
AnnaPolensky, line 47: Do we store the graph as an adjacency list? If so, maybe we could add this information to this comment?

backend/protzilla/step_manager.py
```diff
@@ -0,0 +1,510 @@
...
+    def step_is_terminal(self, step_id: StepID) -> bool:
+        return int(self.graph.out_degree(step_id)) == 0
+
+    def step_is_source(self, step_id: StepID) -> bool:
+        return int(self.graph.in_degree(step_id)) == 0
+
+    def calc_dependencies_met_for_step(self, step_id: StepID) -> bool:
+        """
+        Checks calculation status of all preceding steps in the graph.
+        :return: True iff all preceding steps have been calculated
```
AnnaPolensky, line 136: typo: if

backend/protzilla/step_manager.py
```diff
@@ -0,0 +1,510 @@
...
+        """
+        return all(
+            [
+                step.calculation_status == "complete"
+                for step in self.preceding_steps(step_id)
+            ]
+        )
+
+    def edges_with_exact_data(
```
AnnaPolensky: I feel like the function name should include that we are determining incoming edges (of a specific node).
Edit: I just saw how this function is used and the name makes more sense to me now. It just feel like the docstring implies that the function is more about getting incoming edges than matching some data.

backend/protzilla/step_manager.py
```diff
@@ -0,0 +1,510 @@
...
+    ## Connection management
+    ##
+
+    def connect_steps(self, connection: Connection) -> None:
+        """
+        Connects an output of one source step to an input of another target step.
+        Creates/updates the corresponding link in the graph and sets the handles as edge data.
+
+        :param connection: the connection to establish
+        :return: the target step instance
```
AnnaPolensky: This does not match the return type.

AnnaPolensky (commented):
I only reviewed the step_manager. The code looks really clean and is very well commented. 
However, there are quite a few ToDos and mentions of things that are legacy code and should be handled differently. So just to be sure, do we want to keep it like this or are there some ToDos we need to address before merging to dev?

backend/protzilla/step_manager.py
tE3m: nope, see [here](https://en.wikipedia.org/wiki/If_and_only_if) - iff = gdw

backend/protzilla/step_manager.py
tE3m: We don't store the graph ourselves, the graph is a `networkx.MultiDiGraph` instance, not sure about how they handle it under the hood. Do you think that's important information if we use a library?

backend/protzilla/step_manager.py
AnnaPolensky: Ah, that makes sense

backend/protzilla/step_manager.py
AnnaPolensky: When I was writing this comment I was thinking about the underlying data structure and thought that it is necessary information in order to understand how we work with the graph. But now that I've seen the rest of the code of the StepManager I would agree that it is fine to not mention it since we are not implementing any graph algorithms ourselves.

backend/protzilla/step_manager.py
tE3m: [fixed](https://github.com/cschlaffner/PROTzilla/pull/269/commits/d39da0721ff9d47f5218d604924483ab2ca59cbb)

backend/protzilla/step_manager.py
tE3m: renamed [here](https://github.com/cschlaffner/PROTzilla/pull/269/commits/49401b02c8bc6d7279060879fa894d11cd59db45) to `incoming_edges_for_handle`, do you think that's better?

AnnaPolensky: Yes, sounds good 👍

NeleRiediger:
I've encountered a problem in several runs now where they completely break, forcing me to create a new run. Once, I noticed beforehand that steps could no longer be invalidated after changing values ​​in various fields. After reloading the page, nothing worked at all (see below). It seems that this happens every time I've made changes to the code beforehand. However, restarting the container didn't revive the runs; they were always irretrievably corrupted. I don't recall runs breaking so frequently before, so I wanted to mention it in case there might be a bug somewhere.
Bild 17 03 26 um 12 15

jorisfu:
- All previous runs and workflows not created on this branch (and commonly also created between various commits here) will be broken, this is OK. We need to frequently create new runs until we're finished working on this. This is why we need proper migrations in place to not break runs throughout model changes. The only currently working workflow is the `standard` workflow
- Making changes to the code requires the django dev server to reboot which can take a few seconds. All requests made during boot will return 500 as in your image
- If the django dev server does not boot up, all further requests will return 500. This happens e.g. due to critical python errors like syntax errors or missing imports. See `docker logs -f protzilla-django-1` for the backend logs, these will show such errors

NeleRiediger:
> * All previous runs and workflows not created on this branch (and commonly also created between various commits here) will be broken, this is OK. We need to frequently create new runs until we're finished working on this. This is why we need proper migrations in place to not break runs throughout model changes. The only currently working workflow is the `standard` workflow 

I know, they were all runs I had just created from the standard workflow 

> * Making changes to the code requires the django dev server to reboot which can take a few seconds. All requests made during boot will return 500 as in your image

After one run broke I created a new one to see whether there was just an issue with that one run. Since everything worked fine with the newly created run, I then went back to the previous run, which remained in the state I posted. 

> * If the django dev server does not boot up, all further requests will return 500. This happens e.g. due to critical python errors like syntax errors or missing imports. See `docker logs -f protzilla-django-1` for the backend logs, these will show such errors

Shouldn't those things resolve once the code is corrected and the server is running again?

tE3m:
> I've encountered a problem in several runs now where they completely break, forcing me to create a new run. Once, I noticed beforehand that steps could no longer be invalidated after changing values ​​in various fields. After reloading the page, nothing worked at all (see below). It seems that this happens every time I've made changes to the code beforehand. However, restarting the container didn't revive the runs; they were always irretrievably corrupted. I don't recall runs breaking so frequently before, so I wanted to mention it in case there might be a bug somewhere. Bild 17 03 26 um 12 15

was able to reproduce. This seems to happen when a run is read that contains a model. Apparently, they're still written to the file. investigating now
edit: I think I got it - while outputs are serialized correctly, apparently the input is still just dumped raw.

PR #271, opened 2026-03-02
263 step for alphafold multimer query json generation

### LINKED ISSUE #263: Step for AlphaFold multimer query json generation
## Issue description
We need a new step that lets the user enter a variable number of proteins. Then, a JSON file should be generated that the user can upload to the AlphaFold Server to generate the prediction for these proteins.

## Acceptance Criteria
- [x] Enter a variable number of protein IDs for the prediction.
- [x] For each protein be able to enter the number of copies (>=1).
- [x] Make sure that at least 2 protein ids/2 copies of one protein were entered.
- [x] Use uniprot to get the amino acid sequences of the proteins.
- [x] Generate a json in the AlphaFold json query format (see https://github.com/google-deepmind/alphafold/blob/main/server/README.md)
- [x] Have a new output tab called "Downloads".
- [x] Have a button in the "Downloads"-Tab to download the generated json.

AnnaPolensky:
## Description
fixes #263 
Added a new Output-Tab "Downloads":
grafik
Therefore, introduced a new category of step methods, so that we have calc_method, plot_method and download_method.
Used the new download_method, to add a new step "AlphaFoldMultimerQueryJsonGeneration". User can input a list of uniprot ids and how many copies he wants to have of each protein. One can also add a seed if one wants to use a specific seed. Then a json-File is being generated that can be downloaded via the button in the Downloads-Tab.
grafik

## Changes
backend/protzilla/importing/query_generation.py contains the method for generating the json
backend/protzilla/steps.py contains most of the changes for adding a download_method
frontend/src/components/app/run-screen/run-screen.tsx contains frontend changes for displaying the download-Tab and generating a button for each download.

## Testing

1. Add the query generation step (section importing).
2. Enter protein ids (e.g. I tested with "O43242 O432432" and "2 2", although I believe that that is a query that does not make sense from a biological point of view).
3. Enter a seed or not. (Maybe try both versions.)
4. Calculate step.
5. Go to download-Tab and hit the download button.
6. See that the downloaded file follows this format: https://github.com/google-deepmind/alphafold/blob/main/server/README.md
7. (optional) Go to https://alphafoldserver.com/, continue with your Google-Account and check that you can upload the json:
grafik

Feel free to try different ids and different numbers of copies or try to break the step by putting in different input compared to the specified format.

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**


backend/protzilla/importing/query_generation.py
```diff
@@ -0,0 +1,104 @@
...
+    :return: dict (messages, downloads), downloads contains a dictionary mapping a generated filename
+             to the AlphaFold Multimer query JSON string (wrapped in square brackets as required by AlphaFold server)
+    :raises ValueError: If the number of copies or the model seeds cannot be parsed as integers.
+    :raises requests.exceptions.HTTPError: If fetching a UniProt FASTA sequence fails.
+    """
+    messages = []
+
+    # extract protein_ids and number of copies per id and make sure they have the same length
+    if "," in protein_ids:
+        uniprot_ids = protein_ids.split(",")
```
Elena-kal: Maybe also remove all whitespaces. I am not sure whether my problems came from that...

backend/protzilla/importing/query_generation.py
```diff
@@ -0,0 +1,104 @@
...
+    else:
+        uniprot_ids = protein_ids.split()
+    try:
+        if "," in number_copies:
+            copies_per_id = [int(input) for input in number_copies.split(",")]
```
Elena-kal: On the other hand, this split(",") worked I think

backend/protzilla/importing/query_generation.py
```diff
@@ -0,0 +1,104 @@
...
+            copies_per_id = [int(input) for input in number_copies.split()]
+    except ValueError as e:
+        messages.append(
+            dict(
+                level=logging.ERROR,
+                msg=f"Invalid list of number of copies per id: please provide space-separated integers",
+            )
+        )
+        raise ValueError(
+            "Invalid list of number of copies per id: please provide space-separated integers"
```
Elena-kal:
```suggestion
        msg=f"Invalid list of number of copies per id: please provide space-separated integers"
        messages.append(
            dict(
                level=logging.ERROR,
                msg=msg,
            )
        )
        raise ValueError(
            msg
```

backend/protzilla/importing/query_generation.py
```diff
@@ -0,0 +1,104 @@
...
+                msg=f"For at least one protein id, the number of copies is missing in the input.",
+            )
+        )
+        return dict(messages=messages, downloads={})
+
+    if sum(copies_per_id) < 2:
+        messages.append(
+            dict(
+                level=logging.ERROR,
+                msg=f"Please use the monomer steps, for only validating one protein.",
```
Elena-kal:
```suggestion
                msg=f"Please use the monomer steps to validate only one protein.",
```

backend/protzilla/methods/importing.py
```diff
@@ -607,3 +610,45 @@ def create_form(self):
...
+                NumberField(
+                    name="model_seed",
+                    label="Model seed for AlphaFold",
+                    min=-1,
+                    max=4294967295,
+                    value=-1,
+                ),
+                InfoField(
+                    label="Leave -1 if you want to use a random seed.\n"
+                    "Otherwise enter a seed (=integer between 0 and 4294967295)"
```
Elena-kal:
```suggestion
                    "Otherwise enter a seed (integer between 0 and 4294967295)"
```

backend/tests/main/test_views_helper.py
```diff
@@ -15,6 +15,8 @@ def test_get_all_possible_step_names():
         "FastaImport",
         "AlphaFoldPredictionLoad",
         "CrosslinkingImport",
+        "AlphaFoldMultimerQueryJsonGeneration",
+        "AlphaFoldMultimerQueryJsonGeneration",
```
Elena-kal: Why twice?

backend/protzilla/importing/query_generation.py
```diff
@@ -0,0 +1,104 @@
...
+            {
+                "proteinChain": {
+                    "sequence": amino_acid_sequence,
+                    "count": copies,
+                }
+            }
+        )
+    query_as_string = f"[{json.dumps(query)}]"
+    return dict(
+        messages={},
```
Elena-kal: Why don't we return messages? And even if we want this to be empty, wouldn't we want the messages to be an empty list not a dict?

backend/protzilla/importing/query_generation.py
```diff
@@ -0,0 +1,104 @@
...
+    if len(uniprot_ids) != len(copies_per_id):
+        messages.append(
+            dict(
+                level=logging.ERROR,
+                msg=f"For at least one protein id, the number of copies is missing in the input.",
+            )
+        )
+        return dict(messages=messages, downloads={})
+
+    if sum(copies_per_id) < 2:
```
Elena-kal: Shouldn't we check the individual copies also? We would not want -1 or 0 in their right? Technically we could have 3 copies of one and -1 for the other and this woudl still be false.

backend/protzilla/importing/query_generation.py
```diff
@@ -0,0 +1,104 @@
...
+                "proteinChain": {
+                    "sequence": amino_acid_sequence,
+                    "count": copies,
+                }
+            }
+        )
+    query_as_string = f"[{json.dumps(query)}]"
+    return dict(
+        messages={},
+        downloads={f"prediction_query_{'_'.join(uniprot_ids)}": query_as_string},
```
Elena-kal: this file name could become very long if we use too many uniprot ids. maybe we could truncate it to prevent this.

Elena-kal (changes requested):
One major problem is that inputting the uniprot ids like this: "P69905, P68871" does not work while "P69905 P68871" is fine. Should be fixed by stripping the whitespaces though. 
Another issue I came across is that the step is validated (gets the green checkmark) even though there are error messages. This does not happen in other steps.
I also added a few suggestions to the code but overall the code seems fine. 
I tested a few structures and used the json file for alphafold predictions. This worked very well. I am also very convinced of the downloads tab, I like it a lot and I think it could be useful for other steps as well.

backend/protzilla/importing/query_generation.py
AnnaPolensky: I think, too, that this caused your problem. I originally did not remove the whitespaces because I only added the comma-separated lists to support copy-paste from Excel where no whitespaces would be included. Changed this now.

backend/protzilla/importing/query_generation.py
AnnaPolensky: Because we are having an additional cast to an int here, so the cast kind of removes the whitespaces.

backend/protzilla/importing/query_generation.py
AnnaPolensky: I feel like the user should be capable of not entering a negative amount of copies. However, I added an additional check.

backend/protzilla/importing/query_generation.py
AnnaPolensky: Thanks for pointing out, I added a success message.

backend/protzilla/importing/query_generation.py
AnnaPolensky: I added another input field, the user now enters the filename themselves.

AnnaPolensky:
> One major problem is that inputting the uniprot ids like this: "P69905, P68871" does not work while "P69905 P68871" is fine. Should be fixed by stripping the whitespaces though. Another issue I came across is that the step is validated (gets the green checkmark) even though there are error messages. This does not happen in other steps. I also added a few suggestions to the code but overall the code seems fine. I tested a few structures and used the json file for alphafold predictions. This worked very well. I am also very convinced of the downloads tab, I like it a lot and I think it could be useful for other steps as well.

You can now enter comma- and space-separated ids like "P69905, P68871". Also, changed the validation of the step. Happy to hear that you like the download tab :)

backend/protzilla/methods/importing.py
```diff
@@ -607,3 +610,49 @@ def create_form(self):
...
+
+    output_keys = ["downloads"]
+
+    def create_form(self):
+        return Form(
+            label="AlphaFold Multimer Query JSON Generation",
+            input_fields=[
+                TextField(
+                    name="name",
+                    label="File name and AlphaFold job name for generated query",
```
tE3m, line 640: this label should probably tell the user that the text they enter is only the _stem_ of the filename, since .json is appended automatically

backend/main/views.py
```diff
@@ -624,6 +624,31 @@ def get_step_plots(request):
...
 
 
+def get_step_downloads(request):
+    if request.method == "POST":
+        data = json.loads(request.body)
+        run_name = data.get("run_name")
+
+        run = Run(run_name)
+        if run.current_step is not None:
+            downloads = run.current_downloads.downloads
```
tE3m: this reads rather weirdly - maybe the `Downloads` class should implement `__iter__`, like `Output` does. What does `current.downloads` mean, if we have to access that variable's `downloads` field? Alternatively, `run.current_downloads could` just return `self.current_step.downloads.downloads`

backend/protzilla/importing/query_generation.py
```diff
@@ -0,0 +1,115 @@
...
+        msg = f"Invalid list of number of copies per id: please provide space-separated integers"
+        messages.append(
+            dict(
+                level=logging.ERROR,
+                msg=msg,
+            )
+        )
+        raise ValueError(msg)
+    if len(uniprot_ids) != len(copies_per_id):
+        msg = f"For at least one protein id, the number of copies is missing in the input."
```
tE3m: This is not necessarily true - we only know that the lengths are not equal. There might just as well be more copy counts than protein IDs

tE3m (commented):
some minor adjustments, but good changes overall

backend/protzilla/importing/query_generation.py
tE3m: maybe we could tell the user the length of the lists? They're not exactly easy to read in the current UI

backend/main/views.py
```diff
@@ -682,6 +682,42 @@ def get_step_plots(request):
         )
 
 
+def get_downloads_from_step(request: HttpRequest):
+    """
+    API call. Returns a base64-encoded PNG of a step output to the front-end
+    """
```
jorisfu: fix docstring

jorisfu (changes requested):
Technically fine. Right now we're pretty much limited to "downloads" being basically only JSONs (serialisation both in API and YAML would go haywire if it was anything else). We should refactor this as soon as we add more general downloads, but for now please merge. Mabye add this as a comment somewhere so people don't randomly add byte sequences as "downloads" to their outputs.

PR #288, opened 2026-03-09
feat: add more elaborate selection of the next step after a step was …

### LINKED ISSUE #287: Smart selection of next step
## Issue description
When the user clicks "next" after executing a step, one of the child nodes of the step will be chosen as the next step. Sometimes that's not the most logical choice since a step might be chosen that has preceeding steps that have not been calculated yet but another step might have all its predecessors calculated, so it would be more logical to choose that step.

## Acceptance Criteria
- [x] A step won't be chosen if it has preceeding steps that have not been calculated yet.
- [x] If there are several steps that can be calculated we prefer steps that are children of the current step.
- [x] If none of the child steps of our current step can be calculated immediately, we want to pick a step that can be calculated that is an ancestor of one of our child steps.

AnnaPolensky:
## Description
fixes #287 -> Changed how the next step is picked when clicking the "Next"-Button after a step was calculated.
If a step is a terminal step, the next step is an arbitrary computable step. Otherwise, if there is a computable immediate successor step, it is chosen as the next step. Otherwise, the next step is a computable preceding step of one of the immediate successors.
If all steps are already computed, there won't be a next step.

## Changes
backend/protzilla/step_manager.py: Introduced the function immediate_succeeding_steps and adapted the recommended_next_step_id function.

## Testing
Create a workflow and only navigate between the steps using the "Next"-Button. Check that the behaviour of the "Next"-Button matches the behaviour you would expect based on the acceptance criteria of the issue.

E.g. test the standard workflow with some small modifications:
grafik

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**

### DIFF
--- a/backend/protzilla/step_manager.py
+++ b/backend/protzilla/step_manager.py
@@ -294,6 +294,7 @@ class StepManager:
                     preceeding_step_of_child_step.instance_identifier
                 ):
                     return preceeding_step_of_child_step.instance_identifier
+        return None
 
     def goto_step(self, step_id: StepID) -> None:
         """
--- a/backend/protzilla/step_manager.py
+++ b/backend/protzilla/step_manager.py
@@ -119,6 +119,14 @@ class StepManager:
         descendant_ids = list(nx.descendants(self.graph, step_id))
         return [self.all_steps[step_id] for step_id in descendant_ids]
 
+    def immediate_succeeding_steps(self, step_id: StepID) -> list[Step]:
+        """
+        :param step_id: ID of step of interest
+        :return: List of all immediate successors of the given step (all the child nodes in the graph)
+        """
+        child_ids = list(self.graph.successors(step_id))
+        return [self.all_steps[step_id] for step_id in child_ids]
+
     def step_is_terminal(self, step_id: StepID) -> bool:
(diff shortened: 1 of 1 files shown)


backend/protzilla/step_manager.py
```diff
@@ -251,12 +259,41 @@ def fallback_step_id(self) -> StepID:
...
+            arbitrary_child_step = self.immediate_succeeding_steps(
+                self.current_selected_step_id
+            )[0]
+            for preceeding_step_of_child_step in self.preceding_steps(
+                arbitrary_child_step.instance_identifier
+            ):
+                if _possible_next_step(
+                    preceeding_step_of_child_step.instance_identifier
+                ):
+                    return preceeding_step_of_child_step.instance_identifier
```
jorisfu, line 296: Can we add a `return None` here (at the very end of the function) just in case? Not sure if there is any condition that this doesn't catch and python would return None anyway, but I think it would be more explicit

jorisfu (changes requested):
Very very nice, really like the heuristic. Just one small change pls

NeleRiediger (approved):

PR #290, opened 2026-03-10
289 refactor outputs

### LINKED ISSUE #289: Refactor outputs
## Issue description
Right now there is outputs for outputs of the calc_method and plots for outputs of the plot_method. For some new steps we now have downloads or some visualisation as an output. Therefore, we want to generalise the outputs by adding typing to outputs.

## Acceptance Criteria
- [ ] All steps should still work and produce the expected output.
- [ ] We want to have tables, downloads and visualisations in outputs

jorisfu:
## Description
fixes #289 
Refactored outputs. Outputs is now a dict mapping the output-handle to an OutputItem (consisting of output_type and value). Messages, dataframes, lists and scalars are automatically converted into this format, for all other step outputs OutputItems should be created explicitly. 
The frontend now only displays output-tabs which contain output.

## Changes
Modifications to general output handling in `steps.py`. Serialisation adjustments in `disk_operator.py`. Other changes in other files.

## Testing
Play around with some steps and workflows and tell us which steps are broken so we can fix them yet again

## PR checklist
**Development**
 
**Mergeability**
- [x] 179-branch has been merged into local branch to resolve conflicts
 
**Code review**


backend/protzilla/disk_operator.py
```diff
@@ -43,6 +50,24 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         return True
 
 
+##
+## Custom PyYAML representers/consturctors
```
NeleRiediger: Typo: constructor

NeleRiediger (approved):

backend/protzilla/steps.py
```diff
@@ -459,29 +460,98 @@ def invalidate(self) -> None:
...
+            elif isinstance(value, int):
+                self.output[key] = OutputItem(output_type=OutputType.INT, value=value)
+
+            elif isinstance(value, OutputItem):
+                self.output[key] = value
+
+            else:
+                raise ValueError(
+                    "Outputs must be passed as messages, dataframes, lists or OutputItems"
+                )
```
3dot141592, line 560: I think this may break for outputs like sklearn model, which still occur in e.g. classification.pyandclustering.py (i think elena said something about changing this?).

3dot141592 (commented):
Another thing that breaks: `metadata_column_assignment()` in `backend/protzilla/importing/metadata_import.py` returns `messages=dict()`, but `Output.update()` now only accepts `messages` as a list.

Also: maybe add `bool` and `str` to the supported output types as well? Right now `bool` would fall into the `int`, and plain string outputs do not seem to be handled at all.

backend/protzilla/steps.py
AnnaPolensky: Yes, Elena will take care of this.

AnnaPolensky:
> Another thing that breaks: `metadata_column_assignment()` in `backend/protzilla/importing/metadata_import.py` returns `messages=dict()`, but `Output.update()` now only accepts `messages` as a list.
> 
> Also: maybe add `bool` and `str` to the supported output types as well? Right now `bool` would fall into the `int`, and plain string outputs do not seem to be handled at all.

Since we don't have any steps with bool or str output we are not supporting these types right now. If someone adds a new step with a bool or str output this person should also add this to the supported output types.
We will add something for dealing with messages that are dicts and not lists.

PR #299, opened 2026-03-15
feat: add serialization for non yaml serializable data

Elena-kal:
## Description
We can now read and write non yaml serializable data such as sklearn models on disk.

## Changes
Most changes are in the disk operator.

## Testing
Use any of the clustering or classification steps and check whether they are calculated and protzilla does not crash when reloading the page or when switching to another step and back. (Make sure to build with docker because we have a new library)

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**

### DIFF
--- a/backend/protzilla/disk_operator.py
+++ b/backend/protzilla/disk_operator.py
@@ -117,7 +117,7 @@ class ArtifactOperator:
         with ErrorHandler():
             logger.info(f"Writing artifact to {file_path}")
             file_path.parent.mkdir(parents=True, exist_ok=True)
-            joblib.dump(artifact, file_path)
+            joblib.dump(artifact, file_path, compress=("gzip", 3))
 
 
 RUN_FILE = "run.yaml"
@@ -293,7 +293,7 @@ class DiskOperator:
                     logger.warning(f"Deleting dataframe {file}")
                     file.unlink()
 
-    def clean_artifacts_dir(self, steps: StepManager) -> None:
+    def clean_artifact_dir(self, steps: StepManager) -> None:
         with ErrorHandler():
             if not self.artifact_dir.exists():
                 return
@@ -419,7 +419,8 @@ class DiskOperator:
                         )
                     case OutputType.JOBLIB_ARTIFACT:
                         file_path = (
-                            self.artifact_dir / f"{step.instance_identifier}_{key}.csv"
+                            self.artifact_dir
+                            / f"{step.instance_identifier}_{key}.joblib.gz"
                         )
                         # Only dump if outdated version
                         if self._dump_is_outdated(step, "output"):
@@ -511,16 +512,6 @@ def sanitize_inputs(inputs: dict) -> dict:
             continue
         if utilities.check_is_path(value):
             continue
-        if key == DataKey.PEPTIDE_DF:
-            continue
-        try:
-            yaml.safe_dump(value)
-            sanitized[key] = value
-        except Exception:
-            # not yaml serialisable -> ignore it to prevent errors
-            logger.warning(
-                f"Dropping non-YAML-serializable input {key} of type {type(value)}"
-            )
-            continue
+        sanitized[key] = value
 
     return sanitized
--- a/backend/protzilla/disk_operator.py
+++ b/backend/protzilla/disk_operator.py
@@ -9,6 +9,7 @@ from pathlib import Path
 
 import pandas as pd
 import yaml
+import joblib
 from plotly.io import read_json, write_json
 
 from backend.protzilla.constants.data_types import DataKey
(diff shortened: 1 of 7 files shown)


backend/protzilla/disk_operator.py
```diff
@@ -455,10 +504,23 @@ def sanitize_inputs(inputs: dict) -> dict:
...
+            yaml.safe_dump(value)
+            sanitized[key] = value
+        except Exception:
+            # not yaml serialisable -> ignore it to prevent errors
+            logger.warning(
+                f"Dropping non-YAML-serializable input {key} of type {type(value)}"
+            )
+            continue
+
+    return sanitized
```
jorisfu, line 517: Wait, why do we now use yaml dump for values here?

As far as my understanding of this goes, this function is only used to filter inputs that shouldn't be dumped to disk _again_ as an input, i.e. mostly outputs of other steps, i.e. mostly dataframes. In theory these inputs should be all inputs that are "internal" in some kind of fashion, as the form inputs already get dumped separately.

So IMO there is no need for yaml dumping here, as all of these inputs would be outputs of other steps and as such accessed via handles and already dumped to disk.

@tE3m do you perchance know of a clean way to obtain all inputs that do not fall under form inputs or handle inputs? Do we even have such a scenario?

jorisfu (changes requested):
Really like the implementation, very scalable in the long run as long as you can trust people's run files because [joblib load allows for arbitrary code execution](https://joblib.readthedocs.io/en/latest/persistence.html#persistence). I'm just very unsure about the input sanitation thingy (was bad previously as well sure but we better fix it now lol)

backend/protzilla/disk_operator.py
tE3m: Most steps' inputs should only consist of the form inputs and the handles - any other inputs should be stored in `step.internal_inputs` (a set). I added this attribute in order to hide them from the handles, but could be reused here ig
Then again, idk if there is even any need to dump them at all, since these inputs are silently dynamically loaded, so they'd be recalculated when the step is executed

backend/protzilla/disk_operator.py
```diff
@@ -383,6 +417,17 @@ def _write_output(self, step: Step) -> dict:
                             output_type=OutputType.DATAFRAME,
                             value=str(file_path.relative_to(self.run_dir)),
                         )
+                    case OutputType.JOBLIB_ARTIFACT:
+                        file_path = (
+                            self.artifact_dir / f"{step.instance_identifier}_{key}.csv"
```
tE3m: the dumped artifacts are not in csv format, they should not have the .csv file suffix. See my comment regarding compression - otherwise, there should either be no suffix or sth unique like .artifact

backend/protzilla/disk_operator.py
```diff
@@ -103,6 +104,22 @@ def write(file_path: Path, dataframe: pd.DataFrame):
...
+        with ErrorHandler():
+            logger.info(f"Reading artifact from {file_path}")
+            return joblib.load(file_path)
+
+    @staticmethod
+    def write(file_path: Path, artifact):
+        with ErrorHandler():
+            logger.info(f"Writing artifact to {file_path}")
+            file_path.parent.mkdir(parents=True, exist_ok=True)
+            joblib.dump(artifact, file_path)
```
tE3m: maybe we could compress the artifact? from the [documentation](https://joblib.readthedocs.io/en/stable/generated/joblib.dump.html), we'd then be able to have a  consistent file ending.

backend/protzilla/disk_operator.py
Elena-kal: I mean as long as we know they are always serializable I suppose this will work without dumping them first. This was more a way of checking that they will not cause problems later down the line. But I'll take it out if there is no need for this :))

backend/protzilla/disk_operator.py
```diff
@@ -455,10 +504,23 @@ def sanitize_inputs(inputs: dict) -> dict:
...
-    }
+    sanitized = {}
+
+    for key, value in inputs.items():
+        if isinstance(value, pd.DataFrame):
+            continue
+        if utilities.check_is_path(value):
+            continue
+        if key == DataKey.PEPTIDE_DF:
+            continue
```
tE3m: Why do we check for this key in particular? Shouldn't this already be caught by the instance-check before?

tE3m (changes requested):
Code looks good thus far, just have a look at the comment regarding the filename.
Also, @jorisfu I just had a look at the run.yaml fiel by hand and saw that all form_inputs are also duplicated in the general inputs - should it be that way? Shouldn't those values be reconstructed since they're set as the form fields' values?

tE3m:
oh, and one more thing: classification/clustering tests break because the returned objects are OutputItems. These now need to access that objects .value attribute since it is encapsuled by the OutputItem

backend/protzilla/disk_operator.py
Elena-kal: This was in the sanitize_inputs earlier, now that I took out the safe_dump it should be exactly like before, only written in more readable code... But you are still right, I'll take it out :))

backend/protzilla/disk_operator.py
Elena-kal: Oh, yes you are right, thanks! Probably a copy and paste error 👀

jorisfu:
> I just had a look at the run.yaml fiel by hand and saw that all form_inputs are also duplicated in the general inputs - should it be that way? Shouldn't those values be reconstructed since they're set as the form fields' values?

Yeah no idea what's going on here, didn't touch that part of the disk operator ever. I'll do some testing on the whole dumping-input-again thingy tomorrow. For now I suppose we can merge this into 179.

PR #300, opened 2026-03-16
248 verify runner works

3dot141592:
## Description
Improves the CLI runner for graph-based workflows with multiple import types and aligns runner with the current workflow and output handling.

## Changes
- Extended `runner_cli.py` and `backend/protzilla/runner.py` with additional import arguments (`msfragger`, `diann`, DIA-NN metadata, peptides, evidence, fasta`).
- Updated runner tests and workflow fixtures to match the current graph-based workflow structure.
- Added a bundled `all_steps` workflow plus required test data for broader runner testing.
- Added generic file input mapping via `--file_input_map`, allowing arbitrary file-based step inputs.

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**

## Tests
The failing tests are not caused by changes on this branch


tE3m:
Is this ready to review? Or is there still something missing?

jorisfu:
How would this handle steps like Offline GO Analysis were you can also upload more data? As far as I understand these changes, we only hard-code import arguments for known importing steps, which works for _most_ steps, but is not very scalable in the long run. Is this correct? Also this doesn't work for multiple imports of the same type (e.g. two separate fasta imports don't seem to be supported in the CLI interface)

3dot141592:
> How would this handle steps like Offline GO Analysis were you can also upload more data? As far as I understand these changes, we only hard-code import arguments for known importing steps, which works for _most_ steps, but is not very scalable in the long run. Is this correct? Also this doesn't work for multiple imports of the same type (e.g. two separate fasta imports don't seem to be supported in the CLI interface)

Yes, that is correct, the current approach is not ideal. But is refactoring that really in scope of this PR? Maybe in a follow-up issue?

jorisfu:
> Yes, that is correct, the current approach is not ideal. But is refactoring that really in scope of this PR? Maybe in a follow-up issue?

Well as far as I'm concerned (and as far as we've all been told in the meetings) the first public release on Wednesday should include a working runner and it's fair to assume the runner is only really considered working if it allows for the computation of all workflows also computable using the GUI.

The problem of arbitrary file imports could e.g. by approached by requiring the user to specify a separate file (csv or smth) that describes the file imports for each step where required (i.e. `step, input_field, file_path` tuples). The runner could then validate whether or not that file specifies every necessary import by checking the forms and then execute the workflow if all requirements are met.

That's just my spontaneous idea though, I'm positive you can come up with a cleaner solution for that. Since there is no existing userbase there is no harm in breaking the existing interface here, so go for it

tE3m:
something has gone terribly wrong here with the diff. I can't review 20k changed loc, the last merge commit apparently introduced them

3dot141592:
> something has gone terribly wrong here with the diff. I can't review 20k changed loc, the last merge commit apparently introduced them

https://github.com/cschlaffner/PROTzilla/pull/306

PR #302, opened 2026-03-16
PNG support for GO/GSEA plots

jorisfu:
## Description
Adds frontend and backend support for PNGs artifacts as step outputs encoded as base64. PNGs are shown as "images" in the front-end. Also partially fixes GO/GSEA plots to work with the new system.

## Changes
- Add hidden steps list for users to not instantiate broken steps
- Add Base64 support as step output artifact (to directly work with how both matplotlib and the front-end handle pngs)
- h

NOTE: There is still an issue with selecting gene sets in the GSEA Dot Plot. This step has never properly worked in the first place and I'm a bit tired of fixing stuff other people messed up years before I came here, so this'll have to wait for tomorrow. For now just leave the selection empty (not like you have any other choice)

## Testing
Use the Arbitrary CSV import for your inputs and replicate the tests using the GUI:
- For GO enrichment Bar Plot, use `merged_KEGG_process.csv` from `enrichment_data`
- For GO enrichment Dot Plot, use `Reactome_enrichment_enrichr_2022` from `enrichment_data`
- For GSEA Dot Plot, use `gsea_result_sig_prot.csv` from `enrichment_data`

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**


backend/protzilla/data_integration/di_plots.py
```diff
@@ -37,11 +38,8 @@ def GO_enrichment_bar_plot(
     :type value: str
     :param title: Title of the plot, defaults to ""
     :type title: str, optional
-    :param figsize: Size of the plot, defaults to None and is calculated dynamically if not provided.
-    :type figsize: tuple, optional
 
-    :return: Base64 encoded image of the plot
-    :rtype: bytes
+    :return: the figure as a plotly graph object
```
tE3m, line 42: well, that's not true, is it?

backend/protzilla/data_integration/di_plots.py
```diff
@@ -298,15 +294,16 @@ def gsea_dot_plot(
...
     else:  # remove all Gene_sets that were not selected
+        if not isinstance(gene_sets, list):
+            gene_sets = [gene_sets]
         gsea_df = gsea_df[gsea_df["Term"].str.startswith(tuple(gene_sets))]
 
     if remove_library_names:
         gsea_df["Term"] = gsea_df["Term"].apply(lambda x: x.split("__")[1])
 
     size_y = max((gsea_df[dot_color_value] < cutoff).sum(), 5)
+    print(gsea_df)
```
tE3m: this should probably be removed, right?

backend/protzilla/data_integration/di_plots.py
```diff
@@ -298,15 +294,16 @@ def gsea_dot_plot(
 
     if not gene_sets or gene_sets == "all":
         logger.info("Plotting for all gene set libraries.")
-    if not isinstance(gene_sets, list):
-        gene_sets = [gene_sets]
     else:  # remove all Gene_sets that were not selected
+        if not isinstance(gene_sets, list):
+            gene_sets = [gene_sets]
```
tE3m, line 299: really low-impact, but is there a reason to change this behaviour? it's functionally identical, isn't it?

backend/protzilla/data_integration/di_plots.py
tE3m: nevermind, the rest of the method was collapsed since there were no changes

tE3m (changes requested):
While the plots are successfully displayed, I had an issue with the axis labels: Image
does the axis need to have such a long name? I saw that the prefix could be removed, but still

NeleRiediger (commented):
Testing worked fine and code looks good to me. 
Since the images are very wide, I have to enlarge my window quite a bit to see the actual plot. But I think that's not really part of this issue; rather, we need to enable horizontal scrolling for the plots.

jorisfu:
> does the axis need to have such a long name? I saw that the prefix could be removed, but still

Well, the plot is created by `gsea.dotplot`, so we don't have a lot of choice in adjusting that

backend/protzilla/data_integration/di_plots.py
jorisfu: there was some weird glitch that required me to rewrite this handling a bit, would prefer not touching it again xD

tE3m (approved):
alright, deferring the problem with the plots to the horizontal scrolling issue

PR #303, opened 2026-03-17
301-introduction-of-psm_df

### LINKED ISSUE #301: introduction of psm_df
We are using peptide_df for peptide_dfs and evidence_dfs. Make this clearer by using two different names. evidence_df should be called psm_df (= peptide spectrum match).
The peptide filters should work for both dataframes.

AnnaPolensky:
## Description
fixes #301
PR introduces a psm_df to have clean separation between peptide_df and evidence_df (before both were simply called peptide_df). Added a new handle for the psm_df.
The psm_df should only be used in all the ptm steps and the filters should work for both dataframes. Therefore, I duplicated the filtering steps, so that there are no optional inputs (although we have some code duplication now, but I talked to Chris and he said that this is a valid solution).

## Changes
All the filtering is in data_preprocessing/filter_peptides_or_psm.py (-> before all the functions without a "filter_peptides" or "filter_psm" prefix were in peptide_filter.py)
backend/protzilla/methods/data_preprocessing.py contains the three new filtering steps.

## Testing
Add some peptide filtering, some psm filtering and some ptm steps and try them. (Note: Doing an evidence import takes two or three minutes.)

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**


backend/protzilla/data_analysis/ptm_visualization/ptm_bar_plot.py
```diff
@@ -38,14 +38,14 @@ def get_bar_plot_config_module(
...
-    evidence_df: pd.DataFrame,
+    psm_df: pd.DataFrame,
     evidence_file_q_value_threshold: float,
     fasta_file_path: Path,
     regions_file_path: Path,
     groups_file_path: Path,
 ) -> dict:
     config_module, out_dir = preprocess_files(
-        evidence_df=evidence_df,
+        evidence_df=psm_df,
```
Elena-kal: I am not sure whether this makes sense, but would it be reasonable to rename the input parameter as well -> psm_df=psm_df?
Otherwise at some point we will forget about evidence_df and we'll wonder why we put psm into an evidence field. But I am not sure about this, especially because this is something Hendrik worked on...

backend/protzilla/importing/peptide_import.py
Elena-kal, line 142: Do we still want to call it evidence import? Probably yes, but I still want to put this thopught out. Maybe we can talk to Chris about this, whether leaving the rest like this is fine or whether to at least make it a bit clearer that this is only for ptm steps in protzilla.

backend/protzilla/methods/data_preprocessing.py
```diff
@@ -41,6 +41,11 @@ class OutlierDetectionStep(DataPreprocessingStep, ABC):
     output_keys = [DataKey.PROTEIN_DF]
 
 
+class FilterPsmStep(DataPreprocessingStep, ABC):
+    operation = "filter_psm"
+    output_keys = [DataKey.PSM_DF]
+
+
```
Elena-kal, line 48: I know this is not part of this issue but could we have the same for peptides_df filtering? Would make it more consistent and much cleaner because I really like how you did this.

backend/protzilla/importing/peptide_import.py
Elena-kal: Especially, as we call it PSM in the filtering steps...

tE3m (commented):
It does feel a bit odd to duplicate all the steps. Feels like the "state" pattern could be applied here, but I believe the steps do what they're supposed to.
One thing I think should be changed is the capitalization of PSM: at least all user-facing strings should be PSM and not Psm, since it's an acronym. Apart from that, the code looks good - going to do some functional testing now but from the code, everything looks reasonable

Elena-kal (approved):
Code looks good, just a few comments

backend/protzilla/importing/peptide_import.py
AnnaPolensky: Talked to Chris, we want to keep the step name evidence import since evidence is specific to MaxQuant.

tE3m (approved):

PR #306, opened 2026-03-18
248 verify runner works - the second try >.<

### LINKED ISSUE #248: TODO Verify runner works
## TODO description

We're not certain the runner.py works. This should be checked. Also, verify if there will be collisions with the graph-based workflows.

runner_cli.py

3dot141592:
## Description
Improves the CLI runner for graph-based workflows with multiple import types and aligns runner with the current workflow and output handling.

## Changes
- Extended `runner_cli.py` and `backend/protzilla/runner.py` with additional import arguments (`msfragger`, `diann`, DIA-NN metadata, peptides, evidence, fasta`).
- Updated runner tests and workflow fixtures to match the current graph-based workflow structure.
- Added a bundled `all_steps` workflow plus required test data for broader runner testing.
- Added generic file input mapping via `--file_input_map`, allowing arbitrary file-based step inputs.
- Added automatic export of the input-mapping-file when exporting a workflow

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**


backend/protzilla/data_analysis/plots.py
```diff
@@ -390,6 +390,8 @@ def prot_quant_plot(
         long_to_wide(protein_df) if is_long_format(protein_df) else protein_df
     )
 
+    if not protein_group:
+        protein_group = protein_wide_df.columns[0]
```
tE3m: why? there should never not be a protein group set. This can only happen if there is no incoming protein_df

backend/protzilla/data_preprocessing/filter_samples.py
tE3m, line 1: none of the changes in this file are correct. None of the steps do anything with the peptide_df

backend/protzilla/runner.py
```diff
@@ -63,8 +81,17 @@ def __init__(
...
         )
+        self._validate_file_input_map()
         logging.info(f"Run {self.run_name} created at {self.run.run_path}")
 
+        if (
+            self.run.steps._current_selected_step_id is None
+            and self.run.steps.all_steps
+        ):
+            self.run.steps._current_selected_step_id = (
+                self.run.steps.all_step_ids_toposorted[0]
```
tE3m, line 93: the `run.steps` instance  already has an attribute called `fallback_step_id` - not absolutely necessary to change but I think it'd be better to reuse that here

backend/protzilla/runner.py
```diff
@@ -77,10 +104,15 @@ def __init__(
 
     def compute_workflow(self):
         logging.info("------ computing workflow\n")
-        for i, step in enumerate(self.run.steps.all_steps):
+        ordered_ids = self.run.steps.all_step_ids_toposorted
+        for step_id in ordered_ids:
+            if self.run.steps._current_selected_step_id is None:
+                self.run.steps._current_selected_step_id = step_id
```
tE3m: why? this is literally the sole purpose of `goto_step`

backend/protzilla/runner.py
```diff
@@ -91,46 +123,124 @@ def compute_workflow(self):
 
             if step.calculation_status != "complete":
                 break
```
tE3m, line 123: feels to me like there should be more happening here than just `break` - the run just silently fails here, right?

backend/protzilla/runner.py
```diff
@@ -91,46 +123,124 @@ def compute_workflow(self):
 
             if step.calculation_status != "complete":
```
tE3m, line 122: we mostly stopped passing around actual `Step` instances. While not absolutely necessary, I feel it would be more in line with the rest of the `step_manager` rework to not retrieve the step instance here. the plots and messages can be retrieved via `run.current_plots` and `run.current_messages` - maybe the calculation status could also be added to `run.py`

backend/protzilla/runner.py
```diff
@@ -91,46 +123,124 @@ def compute_workflow(self):
...
+            },
+            "PeptideImport": {
+                "file_path": ("peptides_path", "--peptides_path"),
+            },
+            "EvidenceImport": {
+                "file_path": ("evidence_path", "--evidence_path"),
+            },
+            "FastaImport": {
+                "file_path": ("fasta_path", "--fasta_path"),
+            },
```
tE3m, line 171: I'm not a big fan of this approach, but it works. I believe a [custom action](https://docs.python.org/3/library/argparse.html#action-classes) would be a better fit to handle the old parameters.

runner_cli.py
tE3m, line 1: The newly added arguments are dead on arrival, right? Is this really the way to go? All of the non-input_map parameters should also be marked as deprecated.
Generally speaking, I think a custom action would probably be a _much_ cleaner way of translating the legacy arguments to the new mapping, see my other comment about that

runner_cli.py
```diff
@@ -28,12 +28,41 @@ def args_parser():
...
+    parser.add_argument(
+        "--fasta_path", action="store", help="path to the fasta protein sequence file"
+    )
+    parser.add_argument(
+        "--file_input_map",
+        action="store",
+        help=(
+            "path to a YAML file using step_id -> {field_name: path} for arbitrary "
+            "file-based step inputs"
+        ),
```
tE3m: argparse has [native support](https://docs.python.org/3/library/argparse.html#filetype-objects) for passing files, this is probably more idiomatic

runner_cli.py
```diff
@@ -28,12 +28,41 @@ def args_parser():
...
         "--peptides_path", action="store", help="path to the peptides-dataset"
     )
+    parser.add_argument(
+        "--evidence_path", action="store", help="path to the MaxQuant evidence file"
+    )
+    parser.add_argument(
+        "--fasta_path", action="store", help="path to the fasta protein sequence file"
+    )
+    parser.add_argument(
+        "--file_input_map",
```
tE3m: this would usually be called `file-input-map` in most CLIs

runner_cli.py
tE3m, line 91: I know you didn't touch this, but according to the [docs](https://docs.python.org/3/library/argparse.html#the-namespace-object), the idiomatic way of getting a dict mapping is definitely not a dunder attribute

tE3m, line 92: Possibly out of scope for this PR, but I think it would be much cleaner not to pass _all_ the parameters as arguments of the constructor which just stores them as attributes with the same name. Instead, all required attributes that need some alteration should be parsed via [partial parsing](https://docs.python.org/3/library/argparse.html#partial-parsing) and then passed to the constructor. All the others can just be set by way of passing the `runner` instance as the namespace to the parser, which will set the attributes directly

backend/protzilla/runner.py
tE3m, line 74: This is also extremely silly and should just be handled by setting a default value in the parser.

backend/protzilla/runner.py
```diff
@@ -91,46 +123,124 @@ def compute_workflow(self):
...
+                    raise ValueError(
+                        f"--file_input_map references unknown field '{field_name}' for step '{step_id}'."
+                    )
+                if not isinstance(step.form[field_name], FileInput):
+                    raise ValueError(
+                        f"--file_input_map field '{field_name}' for step '{step_id}' is not a file input."
+                    )
 
     def _perform_current_step(self):
         self.run.current_step.calculate(self.run.steps)
```
tE3m: here, `modify_form` has to be called in order to fill the fields in cases of dynamic filling. otherwise, there is no way of using these steps in the cli

tE3m (changes requested):
The `runner_cli.py` doesn't really work - it didn't before, but e.g. dynamically filled fields are simply empty because `modify_form` isn't called before calculation. While the runner works as such, I think there are still some required and some optional issues left to resolve

jorisfu:
Agree with @tE3m on the required adjustments here. I'd prefer doing a proper review after these issues have been addressed. We especially need `modify_form` calls to properly work with dynamic fill before any further integration testing is done

jorisfu:
Also I'd personally prefer removing the legacy parameters completely and adding custom actions that can override one specific file input each. This would allow for no file map to be passed if all necessary inputs have been given via the cli (i.e. we can still adjust the legacy tests with minimal overhead) but does not bloat the logic too much.

backend/protzilla/data_preprocessing/filter_samples.py
3dot141592: Youre right, this came from all of my merge issues :c

In [this](https://github.com/cschlaffner/PROTzilla/pull/300/changes/3c0ff4847a8f4fc9488e53085ac46b703767b2b0#diff-4c350b53dba2aa81aff536f5253c4f8f6de5d18357ea9b2eafab08df43d7b27b) commit the `peptide_df` parameter was already present in those method signatures. I only added `= None` so it would work with the standard workflow, which obviously didn’t have a `peptide_df` available... At the time, I didnt realize that the dataframe being passed there was already the wrong one :( (which probably should have been obvious after thinking about it for a minute)

ill remove it ^^

backend/protzilla/data_analysis/plots.py
3dot141592: Thats correct, ill remove it.

backend/protzilla/runner.py
3dot141592: What exactly do you find silly about this? I consider it a robust fallback. Except that the code is obviously wrong - It should be:

```
run_name.strip() 
if run_name and run_name.strip()
```

Since `""` is not `None`.

backend/protzilla/runner.py
3dot141592: Since `compute_workflow()` immediately switches to the first topologically sorted step via `self.run.steps.goto_step(step_id)`, the initial selection here isnt very important. So `fallback_step_id` would work, but since it does not necessarily return the topologically first node, I still think it is the worse choice here.

backend/protzilla/runner.py
3dot141592: Yep :) Changed

backend/protzilla/runner.py
3dot141592: Do you mean an explicit abort log here? The step is already marked as failed and its messages are logged right above.

runner_cli.py
3dot141592: Ive changed that, and the other arguments too ^^

backend/protzilla/runner.py
3dot141592: done 👍

jorisfu:
Alright, so what's the current state of this? I'd really like to test this from a user perspective, but we'd need a working state and user guide for that.

Also, we gotta consider HPC environments where we don't have the option of running docker. Since the runner itself only relies on the python requirements, this shouldn't be a huge issue, but needs to be mentioned in the guide. I think it's fair to assume that we don't need to explain how to set up an environment as that differs from machine to machine.

jorisfu:
I cannot save workflows from the GUI:

```
  File "/home/prot/zilla/backend/protzilla/disk_operator.py", line 291, in save_workflow
    step_data[KEYS.STEP_INPUTS] = inputs_to_write
                                  ^^^^^^^^^^^^^^^
NameError: name 'inputs_to_write' is not defined
```

so I cannot test this right now. I believe somthing went wrong when merging with the changes @tE3m did on 179

3dot141592:
> I cannot save workflows from the GUI:
> 
> ```
> Traceback (most recent call last):
>   File "/home/prot/zilla/backend/protzilla/disk_operator.py", line 291, in save_workflow
>     step_data[KEYS.STEP_INPUTS] = inputs_to_write
>                                   ^^^^^^^^^^^^^^^
> NameError: name 'inputs_to_write' is not defined
> ```
> 
> so I cannot test this right now. I believe somthing went wrong when merging with the changes @tE3m did on 179

should work now

jorisfu:
Tested working with both standard workflow and example dataset. Would still change be in favour of removing the legacy inputs completely but this can wait. Let's complete the rest of the release tasks instead today so we're finally done with this.

Approved

jorisfu (approved):

jorisfu:
One more thing though: The file input maps really shouldn't be visible as workflows in the GUI

run_backend_tests.sh
tE3m, line 1: is there a reason to run black inside the container but not the tests?

tE3m (approved):
I also think that the runner/cli interface could use some more love, but it works for now

PR #310, opened 2026-03-20
Fix svm step

jorisfu:
## Description
The svm step works now. Before it had parameters the user could not set and did not run through because of possible NaN values. 

## Changes
I needed variable input fields because the parameter "class_weights" is a dictionary with inputs depending on other inputs. I therefore needed to change the step class

## Testing
I have a workflow in our next cloud folder. There is a workflow with a svm step. You can also create one yourself. But there are a few preprocessing steps needed I believe, however I am not sure which and I just use the standard workflow and squeze svm in between steps. (One more thing: When selecting no positve class the step throws errors on my machine. I am currently talking to Chris about it, but this is most likely a problem of how I set my parameters and less the svm step itself.)

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**


backend/protzilla/data_analysis/classification_helper.py
```diff
@@ -217,11 +220,11 @@ def perform_train_test_split(
     test_size=0.2,
     random_state=42,
     shuffle=True,
-    split_stratify="yes",
+    split_stratify=True,
 ):
     # by default this contains already filtered samples from metadata, we need to remove those
     labels_df = labels_df[labels_df.index.isin(input_df.index)]
-    split_stratify = labels_df if split_stratify == "yes" else None
+    split_stratify = labels_df if split_stratify else None
```
AnnaPolensky: You do not need to change it, but I'm not sure whether I'm a big fan of storing a df in split_stratify after split_stratify was a boolean. Personally, I think I would prefer something like this:
```suggestion
    return train_test_split(
        input_df,
        labels_df,
        test_size=test_size,
        random_state=random_state,
        shuffle=shuffle,
        stratify=labels_df if split_stratify else None,
    )
```

backend/protzilla/methods/data_analysis.py
```diff
@@ -1677,10 +1677,53 @@ class ClassificationSVM(ClassificationStep):
...
+            else:
+                form.add_field(
+                    FloatField(
+                        name=field_name,
+                        label=f"Weight of {cl.value}",
+                        min=0.0,
+                        value=1.0,
+                    )
+                )
+
```
AnnaPolensky, line 1713: Instead of hiding all fields and then showing all fields that are in classes and creating new fields for all the fields that are in classes but do not already exist, could we just not create any of the weight-fields beforehand and then directly create a field for each element in classes?

AnnaPolensky (approved):
Do we want any tests for the changes?

backend/protzilla/methods/data_analysis.py
Elena-kal: I am not sure what you mean :(( I did this because we cannot delete fields but if we switch the column we need different classes and therefore different class weights. I we switch back we want them to be visible again. I think it would be unnecessary to create all fields beforehand as this could get quite large if we have too many columns and more than 5 classes.

Elena-kal:
> Do we want any tests for the changes?

There were no tests for svm - so I didn't adjust anything. I guess we could create tests but I do not think this is part of this issue.

backend/protzilla/methods/data_analysis.py
AnnaPolensky: I was not aware that we could not delete fields and that the visible fields depend on the column and we could switch between different columns. Makes sense now.

AnnaPolensky:
> > Do we want any tests for the changes?
> 
> There were no tests for svm - so I didn't adjust anything. I guess we could create tests but I do not think this is part of this issue.

I was primarily thinking about the changes in form.py. You only added two small functions, but I feel like at some point it could happen that someone else might want to extend or slightly change the functionality and then it might be useful to have a test making sure that your functionality is still there, so that svm does not break then. But, I'm unsure how likely this is.

Elena-kal:
> > > Do we want any tests for the changes?
> > 
> > 
> > There were no tests for svm - so I didn't adjust anything. I guess we could create tests but I do not think this is part of this issue.
> 
> I was primarily thinking about the changes in form.py. You only added two small functions, but I feel like at some point it could happen that someone else might want to extend or slightly change the functionality and then it might be useful to have a test making sure that your functionality is still there, so that svm does not break then. But, I'm unsure how likely this is.

Makes sense, but I don't think we even have tests for forms...

PR #317, opened 2026-03-25
v1.0.0 merge into main

tE3m:


tE3m:
ok nvm, squash-merging was the wrong call here. this loses the commit history, that's really bad. reverting this now

PR #320, opened 2026-03-26
Changes output of example dataset import to psm_df

hendraet:
## Description

Changes the output of the example dataset import from peptide_df to psm_df. Was an oversight in a previous attempt to separate the namings for peptides and evidence dataframes.

## Testing

Create a run with the example dataset import and verify that the output is a psm_df (and works with steps that require psm_df, such as the ptm visualizations)


jorisfu (approved):
Works but please change base branch to `dev` before merging

PR #321, opened 2026-03-27
266 file picker in forms

### LINKED ISSUE #266: File picker for imports in step forms
## Issue description
When trying to import a file into a run via a step the file browser only lets the user choose files from the correct file type.
Hint: This is already implemented for imports in settings and for run imports. It just needs to be extended to also work for imports in steps.

## Acceptance Criteria
- [ ] when selecting a file for an import the user can only choose files with a specific file type specified by the software

Elena-kal:
## Description
fixes #266 
Before the user could choose any file for the file import although we had a file picker implemented for the frontend but did not use it in the backend. The user should now only be able to import the file types we want them to. (Well not really, there is a way around it - but at least it is visible which file types are expected.)

## Changes
One little line in forms.py and everywhere where we use the File Import Field.

## Testing
Create a run and check out any step that requires you to upload a file.

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**

### DIFF
--- a/backend/protzilla/methods/importing.py
+++ b/backend/protzilla/methods/importing.py
@@ -60,7 +60,12 @@ class ArbitraryCSVImport(ImportingStep):
         return Form(
             label="Arbitrary CSV Import",
             input_fields=[
-                FileInput(name="file_path", label="CSV file", value=None, accept=".csv")
+                FileInput(
+                    name="file_path",
+                    label="CSV file",
+                    value=None,
+                    accept=".csv",
+                )
             ],
         )
 
@@ -87,7 +92,7 @@ class MaxQuantImport(ImportingStep):
                     name="file_path",
                     label="MaxQuant intensities file (proteinGroups.txt)",
                     value=None,
-                    accept=".txt",
+                    accept=".txt,.tsv,.csv",
                 ),
                 DropdownField(
                     name="intensity_name",
@@ -132,7 +137,7 @@ class DiannImport(ImportingStep):
                     name="file_path",
                     label="DIA-NN intensities file (*.pg_matrix.tsv)",
                     value=None,
-                    accept=".tsv",
+                    accept=".txt,.tsv,.csv",
                 ),
                 CheckboxField(
                     name="map_to_uniprot",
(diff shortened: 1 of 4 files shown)


backend/protzilla/methods/importing.py
```diff
@@ -91,6 +92,7 @@ def create_form(self):
                     name="file_path",
                     label="MaxQuant intensities file (proteinGroups.txt)",
                     value=None,
+                    accept=".txt,.tsv,.csv",
```
tE3m: This is a tricky one. While any of `csv`, `tsv`, `txt` are possible, only `txt` makes sense, right? MQ will always output the file as `proteingroups.txt`, so does it make sense to allow the other ones?

tE3m: Same goes for all the other tool-specific imports (DIA-NN, MSFragger, MQ peptides/evidence)

tE3m (commented):
Good changes! Not sure about all allowed suffixes though - take a look :)

backend/protzilla/methods/importing.py
hendraet: While you are correct, I wouldn't be overly strict here. Rather allow more file types that could be parsed and not unnecessarily restrict the user. 
I could imagine there's at least one biologist out there who has MaxQuant-like protein data in a file that doesn't end in .txt.

tE3m: I get that and I wouldn't suggest restricting it further if this implementation actually didn't _allow_ selecting files that don't adhere to the accept specification. At least on all machines I tested, the file picker allows ignoring the accept spec. I'm also fine with leaving as is.

hendraet: Yeah sorry, I didn't phrase it properly. I know that it also allows other file types, but at least on my OS, it requires an extra click. This is at most a minor inconvenience, so it wouldn't be a big deal to implement your proposal.

Elena-kal: I made it a bit stricter. Some definitely take and expects tsv, here I only took out the csv because txt seems to be some kind of standard everywhere. For peptide, MaxQuant and Evidence I restricted it to only txt. What do you think?

tE3m (approved):
lgtm 👍

PR #331, opened 2026-04-05
Reimplement z-scorce significance calculation

hendraet:
## Description

This PR refactors the z-score significance calculation that is part of the AML paper reproduction. There were still some minor mismatches in the numbers that the original paper reported (Supplementary 1) and our calculation. I dug in a bit and found the original implementation documented in this paper https://pubs.acs.org/doi/10.1021/pr1009977.

Our preprocessing still seems a bit off, so that we don't match the significant proteins of the original paper perfectly, but just doing the calculation on the FCs of the original paper, we are now pretty close. Also, we achieve good overlap in proteins when setting the z-score cutoff to 0.05 (used in the AML paper) instead of 0.07 (which felt more like a magic number).

## Testing
 
Run the example workflow, including the differential expression step, in its default configuration. Also, the newly added test should reproduce the numbers from the original [supplementary material](https://www.mdpi.com/2072-6694/12/3/709/s1) quite nicely.

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**


jorisfu (approved):
Tested working and looks good

PR #344, opened 2026-04-13
Fix Plot Customizability

### LINKED ISSUE #139: Improve plot export
## Issue description

Currently, the plot export is not working as intended, as it is impossible to export some figures in a way that they look like in the plot output. Additionally, it is sometimes not even possible to read all labels because there is an overlap of labels and other figure elements.

Most prominent examples are basically all outputs of the PTM visualization steps:
Image

Also, exporting figures that include subplots is not really possible because it will introduce large margins that result in whitespaces while plots are squished. Originally, the protein coverage plot was a single figure with many subplots, but I had to split it into multiple plots to avoid this and general display issues:

Image

## Acceptance Criteria
- [ ] Figures can be exported in a way that they are actually nicely readable and visually appealing. It should at least be possible to export the images so that they look like in the plot output.
- [ ] It is possible to export figures that contain multiple subplots without losing a lot of space because of unnecessary margins.

hendraet:
## Description
fixes #139

- Changes the scaling of the plot preview in the plot download modal so that it leaves room for more actual customizability. Before, the preview was too small, and details were lost. The export just upscaled the tiny version, resulting in no gain of detail, even when adjusting export width and height.
- Adds some more guard rails when mutating plots

## Testing

The clustergram from the example workflow serves as a nice test because it was not possible to export it with sufficient detail before these fixes. 

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**

### DIFF
--- a/frontend/src/components/app/settings/plot-settings/usePlotSettings.ts
+++ b/frontend/src/components/app/settings/plot-settings/usePlotSettings.ts
@@ -61,11 +61,16 @@ export const usePlotSettings = (isOpen?: boolean) => {
     textSize: 0,
   });
 
-  const loadSettings = async (templateName: string) => {
-    const response = await callApiWithParameters("load_settings", {
-      templateName: templateName,
-    });
-    if (response) {
+  const loadSettings = async (templateName: string): Promise<PlotSettings | null> => {
+    setIsLoading(true);
+    try {
+      const response = await callApiWithParameters("load_settings", {
+        templateName: templateName,
+      });
+      if (!response) {
+        return null;
+      }
+
       const loadedSettings: PlotSettings = {
         fileFormat: response.file_format,
         width: response.width,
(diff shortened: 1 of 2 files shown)


frontend/src/components/app/settings/plot-settings/usePlotSettings.ts
```diff
@@ -148,41 +173,48 @@ export const usePlotSettings = (isOpen?: boolean) => {
...
     }
   };
 
-  // Fixed width for containing the correct ratio of width & height
-  const basePlotWidth = 400;
   const ptToInch = 1 / 72;
   const inchToMm = 25.4;
-  const dpi = 300;
+  const exportDpi = 300;
+  const screenDpi = 96;
```
jorisfu, line 196: where does this come from?

jorisfu (approved):
Just one small question but code looks good and works as expected with the clustergram

frontend/src/components/app/settings/plot-settings/usePlotSettings.ts
hendraet: It's a commonly used assumption for screen dpi. It might not be ideal anymore, but I'm not sure if it's worth more engineering effort.

PR #346, opened 2026-04-13
manage output tab selection externally

### LINKED ISSUE #345: BUG: Uninitialised output tabs trigger output tab switching when loaded
# What is the issue?
When some outputs load longer, the then appearing output/plot tab may be selected automatically, switching between outputs without any user interaction

# How to reproduce the issue?
Load a sizeable dataset and perform Transformation: Log. First, the tables without content will be shown, then the plot appears and the output tab automatically switched to Plots

The selected output tab does never change if the user doesn't intent to change it

jorisfu:
## Description
fixes #345 
Output tabs do not randomly change to previously unselected output types if run data changes

## Changes
The selection of the output component to display is now managed externally in the run screen. Whenever you select e.g. "Tables", all navigation between steps and state reloads will also lead to Tables being displayed (if available), no fallback to plots happens. This also avoids some issues with lazy loading.

Slight caveat: Long loading times of certain output types (e.g. large plots) may still lead to output switching if the previously selected component was that output type but another one gets loaded prior. This should generally be fixed by implementing lazy loading throughout all output types and is out of scope for this fix.

## Testing
Load a sizeable dataset and perform Transformation: Log. Previously, the semi-loaded tables would be shown and then the output would switch to Plots as soon as they are loaded. Now, per default the table is selected and does not switch. If you switch to plots and recalculate, the plots should be shown again. Same with tables, no permanent switching to previously not selected outputs

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**

### DIFF
--- a/frontend/src/components/app/run-screen/run-screen.tsx
+++ b/frontend/src/components/app/run-screen/run-screen.tsx
@@ -98,6 +98,8 @@ export const RunScreen: React.FC = () => {
 
   const [isDownloadModalOpen, openDownloadModal, closeDownloadModal] = useToggleableState(false);
 
+  const [selectedOutputTab, setSelectedOutputTab] = useState<SwitchComponent["name"]>("Tables");
+
   const getFooterMessage = () => {
     const currentTimestamp = new Date();
     const currentHour =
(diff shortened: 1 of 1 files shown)


jorisfu:
Already fixed with #268

PR #270, opened 2026-03-01
267 cl validation for multimer structures

Elena-kal:
## Description
fixes #267 
Users can now validate multimer structures, the same way they can validate monomer structures.

## Changes
Step was added in data_analysis.py, the logic behind the validation can be found in crosslinking_validation.py. I mostly refactored and extended the existing code. I also added tests.

## Testing
Import the crosslinking data, download the files from this [folder](https://nextcloud.hpi.de/s/PGn2SyCPYHm8JQT) (password: protzilla!). You can use these for the multimer structure prediction upload. The two proteins are P62191 and P62195-2. Validating them should result in 5 crosslinks. (This only tests that everything works. These two proteins have crosslinks in the crosslinking data but I do not know whether they are of any relevance and have a specific structure together. I simply used the first proteins from the inter crosslinks I could find.) Feel free to validate other protein structure predictions in the crosslinking data.

## PR checklist
**Development**
- [x] If necessary, I have updated the documentation (README, docstrings, etc.)
- [x] If necessary, I have created / updated tests.
 
**Mergeability**
- [x] main-branch has been merged into local branch to resolve conflicts
- [x] The tests and linter have passed AFTER local merge
- [x] The backend code has been formatted with `black`
- [x] The frontend code has been formatted with `pnpm format` and checked with `pnpm lint`
 
**Code review**
- [x] I have self-reviewed my code.
- [x] At least one other developer reviewed and approved the changes

PR #357, opened 2026-04-21
351 correct multimer validation

### LINKED ISSUE #351: Correction of multimer validation
## Issue description
Currently, multimer validation does not take into account that if there are several of the same proteins within the structure it is unclear where the crosslink binds. It could bind between all combinations of them.
We therefore want to duplicate the crosslinks for each possible combination. We also want to add the chain_ids to the crosslinking df.

## Acceptance Criteria
- [ ] Chain Id 1 and Chain Id 2 in crosslinking df
- [ ] All possible combinations of crosslinks between different chains are validated

Elena-kal:
## Description
fixes #351 
Before, the multimer crosslinking validation was wrong because it did not account for the chain id or the protein id. Now it looks at both.

## Changes
 Because there is no protein id in the cif and only their labels going from 1 to the number of different chains I needed to add a mapping from the protein ids to the labels. One can create a mapping like this by looking into the job request json for the alphafold prediction. That is why I needed to add this as well and changed the importing.

## Testing
Look at the previous mulitmer PR ([Multimer validation PR](https://github.com/cschlaffner/PROTzilla/pull/270)) and create a workflow like this again. Also try uploading the multimer validation in different places (settings and in the workflow) as we have a new form of input as well.
Also check the monomer validation by creating a workflow described by Anna in this PR:  [Monomer validation PR](https://github.com/cschlaffner/PROTzilla/pull/212)
Here is the link to the nextcloud folder where you can find data to upload for the multimer validation: [https://nextcloud.hpi.de/s/PGn2SyCPYHm8JQT](https://nextcloud.hpi.de/s/PGn2SyCPYHm8JQT)

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**


backend/protzilla/data_analysis/crosslinking_validation.py
```diff
@@ -256,64 +268,269 @@ def get_crosslink_positions_in_protein(
...
+    for protein_id in structures_to_validate:
+        chains_per_protein[protein_id] = get_chains(
+            cif_df=cif_df,
+            valid_ids=valid_ids,
+            protein_id=protein_id,
+            id_column_name=id_column_name,
         )
 
-    relevant_crosslinks_df = all_crosslinks_df[mask].copy()
+    relevant_crosslinks_df = expand_crosslinks_to_chain_combinations(
```
NeleRiediger, line 537: Do we need to call this, if we are dealing with a monomer? I don't no wether it is worth adjusting, just feels weird to always call it.

NeleRiediger (approved):
Testing worked as expected and code looks fine as well. 
I made a small comment, but I think the code is fine either way, so code changes aren't necessarily required.

backend/protzilla/data_analysis/crosslinking_validation.py
Elena-kal: Like we talked about, I know it is irrelevant for monomers, however we would need to split monomer and multimer validation even further.

backend/protzilla/data_analysis/crosslinking_validation.py
```diff
@@ -256,64 +268,269 @@ def get_crosslink_positions_in_protein(
...
+    :return: the Protein ID that matches the sequence. (Should only be one, therefore we take the first one)
+    """
+    matching_rows = amino_acid_sequences_df[
+        amino_acid_sequences_df["Protein Sequence"] == target_sequence
+    ]
+    if not matching_rows.empty:
+        return matching_rows["Protein ID"].iloc[0]
     else:
-        raise ValueError("Metadata must contain 'uniprot_ids' or 'uniprot_accession'.")
+        return None
```
3dot141592, line 399: Have you considered the case where multiple protein IDs share the same sequence? In that case this seems to pick the first match and could assign all copies to the same protein ID.

3dot141592 (commented):
I just had one question, everything else looks good :)

backend/protzilla/data_analysis/crosslinking_validation.py
Elena-kal: I think because protein IDs should be unique to their sequence that should be fine. We can expect each sequence to have only one protein ID.

PR #380, opened 2026-04-27
Invalidate plot method if calculation does not return a plot

### LINKED ISSUE #377: BUG: false validation of Bar plot forGO enrichment analysis
# What is the issue?
For the step: Bar plot for GO enrichment analysis, if no component is selected and the step is calculated you get an error, that at least one component needs to be selected. Everything correct so far. However the step still gets its little green validation mark , even though it correctly threw an error. 

# How to reproduce the issue?
Use the standard workflow and calculate everything up to this step. For the Bar plot for GO enrichment analysis step make sure that no component is selected and press calculate. You should see the error popping up but also the validation mark turning green. 

The error should still appear in this scenario as it is absolutely correct. However the little validation mark should turn red and not green in this case.

jorisfu:
## Description
fixes #377 

Amends the plot output handling to raise an error if no plots were generated. Prior to this, plot steps would be marked as valid even though the calculation failed

## Testing
Use the standard workflow and calculate everything up to this step. For the Bar plot for GO enrichment analysis step make sure that no component is selected and press calculate. You should see the error popping up but also the validation mark turning red.

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**

### DIFF
--- a/backend/protzilla/steps.py
+++ b/backend/protzilla/steps.py
@@ -325,6 +325,9 @@ class Step(ABC):
         else:
             plots = outputs
 
+        if not plots:
+            raise ValueError("Output of plot method is empty.")
+
         self.plots = Plots(plots)
 
     def handle_messages(self, outputs: dict) -> None:


NeleRiediger (approved):
I think this fixes the problem well enough and it works as expected

PR #390, opened 2026-04-29
206 dimensionality reduction on proteins

### LINKED ISSUE #206: Dimensionality Reduction on Proteins
## Issue description

Currently, dimensionality can only be to samples It would be nice if the user could choose to apply it to the proteins as well so that they can be compared nicely.

## Acceptance Criteria
- [x] Dimensionality reduction steps (currently UMAP, t-SNE) have an option to select if sample or protein dimension should be reduced
- [x] The results can be visualized in the scatter plot step
- [x] Written tests

ferbsx:
## Description
fixes #206 
Added dimensionality reduction on protein level. 

## Changes

Before: dimensionality reduction only done on sample level (per default, not customisable)
Now: added drop down to choose which level the dimensionality reduction should be applied on.

Options: 
- Sample [default]
- Protein ID

Adapted scatter plot requirements to allow plotting of the results based on Protein ID.

## Testing

Former tests are updated to pass with the new structure. New tests added for all functionalities.

## PR checklist
**Development**
 
**Mergeability**
- [x] The tests and linter have passed AFTER local merge [only GSEA tests are failing which has been discussed]
 
**Code review**

### DIFF
--- a/backend/protzilla/data_analysis/plots.py
+++ b/backend/protzilla/data_analysis/plots.py
@@ -38,6 +38,7 @@ def scatter_plot(
     input_df: pd.DataFrame,
     metadata_df: pd.DataFrame | None = None,
     metadata_column: str | None = None,
+    sample_name: str = "Sample",
 ) -> dict:
     """
     Function to create a scatter plot from data.
@@ -49,6 +50,7 @@ def scatter_plot(
     :param metadata_column: the name of the column in `metadata_df` that contains the
         group information for each sample. This parameter is required if `metadata_df`
         is provided.
+    :param sample_name: the name of the sample column, should be the same name for the input_df and the metadata_df
 
     :return: returns a dictionary containing a list with a plotly figure and/or a list of messages
     """
@@ -58,19 +60,24 @@ def scatter_plot(
                 "The column selected for annotation is not present in the corresponding metadata dataframe.",
             )
 
+    if sample_name not in input_df.columns:
+        raise ValueError(
+            f"The column {sample_name} selected for annotation is not present in the corresponding input dataframe.",
+        )
+
     intensity_df = input_df.copy()
     if isinstance(metadata_df, pd.DataFrame):
         intensity_df = pd.merge(
             intensity_df,
-            metadata_df[["Sample", metadata_column]],
-            on="Sample",
+            metadata_df[[sample_name, metadata_column]],
+            on=sample_name,
             how="left",
         )
     else:
         # Mock a metadata column here so that we can treat dfs with and without metadata the same way
         metadata_column = "mock_metadata_column"
         intensity_df[metadata_column] = None
-    intensity_df = intensity_df.drop(columns="Sample")
+    intensity_df = intensity_df.drop(columns=sample_name)
 
     color_col = (
         metadata_column if intensity_df[metadata_column].notnull().any() else None
--- a/backend/protzilla/methods/data_analysis.py
+++ b/backend/protzilla/methods/data_analysis.py
@@ -904,6 +904,12 @@ class PlotScatterPlot(DataAnalysisPlotStep):
                     name="metadata_column",
                     label="Choose the column of the metadata dataframe that should be used for coloring",
                 ),
+                DropdownField(
+                    name="sample_name",
+                    label="Choose the column that contains the sample information",
+                    value=DimensionReductionValue.sample.value,
+                    options=DimensionReductionValue,
+                ),
             ],
         )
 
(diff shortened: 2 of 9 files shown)


backend/protzilla/methods/data_analysis.py
```diff
@@ -899,6 +904,12 @@ def create_form(self):
                     name="metadata_column",
                     label="Choose the column of the metadata dataframe that should be used for coloring",
                 ),
+                DropdownField(
+                    name="sample_name",
+                    label="Choose the column that contains the sample information",
```
hendraet, line 909: Calling it sample and the variable `sample_name` might confuse the user because it is too close to the actual "Sample" column and it might not be clear that "Protein ID" could also be a valid "sample column" in this case.

backend/protzilla/data_analysis/plots.py
hendraet, line 57: If you want to plot proteins instead of samples, currently, you cannot connect a metadata dataframe - otherwise, this if and its error are triggered. I feel like we should make it more transparent to the users that they shouldn't connect metadata in this case

backend/protzilla/utilities/transform_dfs.py
```diff
@@ -20,9 +33,7 @@ def long_to_wide(intensity_df: pd.DataFrame, value_name: str | None = None):
     values_name = (
         default_intensity_column(intensity_df) if value_name is None else value_name
     )
-    return pd.pivot(
-        intensity_df, index="Sample", columns="Protein ID", values=values_name
-    )
+    return pd.pivot(intensity_df, index=index, columns=columns, values=values_name)
```
hendraet, line 36: Maybe we should add some guardrails to make sure that `index` and `columns` are not the same, because this will lead to a cryptic error message. (This should ideally never happen, but if I had a penny for every time I didn't check assumptions because they could never possibly happen...)

backend/protzilla/data_analysis/plots.py
```diff
@@ -38,6 +38,7 @@ def scatter_plot(
     input_df: pd.DataFrame,
     metadata_df: pd.DataFrame | None = None,
     metadata_column: str | None = None,
+    sample_name: str = "Sample",
```
hendraet, line 41: See comment above, I feel like `sample_name` is misleading. 
However, it seems like it is only for metadata processing. Since metadata is enforced to include a "Sample" column (different problem), and using "Protein ID" together with metadata will lead to errors anyway, one could probably also revert the whole `sample_name` completely.

backend/tests/protzilla/data_analysis/test_dimension_reduction.py
```diff
@@ -250,6 +252,85 @@ def test_tsne_scatter_plot_integration(
     )
 
 
+@pytest.mark.parametrize(
+    "df_name,n_components,method",
+    [
+        ("dimension_reduction_df", 2, TSNEMethod.exact.value),
+        ("dimension_reduction_four_proteins_df", 2, TSNEMethod.exact.value),
```
hendraet, line 259: Would like to have all of these tests also test 3 components

backend/protzilla/data_analysis/plots.py
```diff
@@ -58,19 +60,24 @@ def scatter_plot(
                 "The column selected for annotation is not present in the corresponding metadata dataframe.",
             )
 
+    if sample_name not in input_df.columns:
```
hendraet, line 63: Would like to see a test that checks the raising of this `ValueError`

backend/tests/protzilla/data_analysis/test_dimension_reduction.py
```diff
@@ -271,6 +352,26 @@ def test_umap(dimension_reduction_df, n_components):
         )
 
 
+@pytest.mark.parametrize("n_components", [2])
```
hendraet, line 355: Either use more than one value for `n_compontents` or don't parametrize at all

hendraet (commented):
Works as intended, but needs some minor refinements.

However, in its current state, the PR might not be as useful as imagined because I underspecified the issue. What would be needed to increase usefulness:
- Hover annotations for each data point in the scatter plot. Currently, we lose all information about Samples/Protein IDs in the scatter plot, which would be especially helpful in the protein case. (Could be an easy fix, currently we just deliberately exclude this information from the plot, but it is passed to the function)
- Having no metadata available to color proteins in a scatter plot is also less than ideal. However, we currently cannot import any metadata that does not include a `Sample` column, so one could probably only do it in a hacky way.

Issue #371, opened 2026-04-24
BUG: Fix display of vertical lines in CL plots


AnnaPolensky:
For multimere validation: The vertical line for the CL length is always displayed at 0 even if it is not 0?

PR #395, opened 2026-05-04
Improve CL validation plots

### LINKED ISSUE #372: Add logarithmic scaling to CL plots

AnnaPolensky:
## Description
Improves CL validation plots and addresses #371 and #372 
- Add wrapping with line breaks for too long titles. (Before they weren't fully displayed, they just stopped.) -> backend/protzilla/data_preprocessing/plots.py
- Add a split diagram with linear and logarithmic scaling. The split is the CL length or if existent the CL length + allowed_upper_deviation. Using plotly's histogram function for logarithmic histograms did not work out very well, which is why we decided to use numpy to create logarithmic bins (size of the bins grows according to logarithmic scale) -> backend/protzilla/data_analysis/crosslinking_validation.py
- Add tick values for the vertical lines (they represent the CL length and allowed deviations). -> _get_tick_values_with_lines in backend/protzilla/data_analysis/crosslinking_validation.py
- Add a bit more information to the legends. (concrete values for the length and allowed deviations)
grafik

## Testing
Do a monomer and multimer validation. 
Try different lengths and upper and lower bounds on accepted deviation (test only upper, only lower, or both at the same time).

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**

### DIFF
--- a/backend/protzilla/data_analysis/crosslinking_validation.py
+++ b/backend/protzilla/data_analysis/crosslinking_validation.py
@@ -1034,8 +1034,8 @@ def create_cl_validation_histogram(
     distances_valid.dropna(inplace=True)
     distances_invalid.dropna(inplace=True)
 
-    min_distance: float = np.min([distances_valid.min(), distances_invalid.min()])
-    max_distance: float = np.max([distances_valid.max(), distances_invalid.max()])
+    min_distance: float = np.nanmin([distances_valid.min(), distances_invalid.min()])
+    max_distance: float = np.nanmax([distances_valid.max(), distances_invalid.max()])
 
     fig = make_subplots(
         rows=1, cols=2,
--- a/backend/protzilla/data_analysis/crosslinking_validation.py
+++ b/backend/protzilla/data_analysis/crosslinking_validation.py
@@ -9,6 +9,7 @@ import re
 import logging
 
 from pandas.io.stata import stata_epoch
+import plotly.graph_objects as go
 from plotly.graph_objects import Figure
 
 from backend.protzilla.data_preprocessing.plots import (
@@ -22,6 +23,17 @@ from backend.protzilla.data_analysis.plots import (
 from backend.protzilla.steps import OutputItem, OutputType
 from backend.protzilla.data_preprocessing.plots_helper import millify
 
+import textwrap
+
+from plotly.subplots import make_subplots
+
+from backend.protzilla.data_preprocessing.plots_helper import generate_tics
+from backend.protzilla.utilities.utilities import default_intensity_column
+from backend.protzilla.constants.colors import (
+    PLOT_COLOR_SEQUENCE,
+    PLOT_PRIMARY_COLOR,
+    PLOT_SECONDARY_COLOR,
+)
 
 def get_reactive_atom_of_amino_acid_residue(amino_acid_type: str) -> str:
     """
@@ -781,7 +793,7 @@ def diagrams_of_crosslinking_validation_data(
             )
             + 1
         )
-        histogram = create_histograms(
+        histogram = create_cl_validation_histogram(
             dataframe_a=df_valid,
             dataframe_b=df_invalid,
             name_a=f"Valid Crosslinks (intra: {valid_intra}, inter: {valid_inter})",
@@ -790,11 +802,8 @@ def diagrams_of_crosslinking_validation_data(
             x_title="Distance (Å)",
             y_title="Count",
             overlay=True,
-            visual_transformation="linear",
             relevant_column_a="alphafold_distance",
             relevant_column_b="alphafold_distance",
-            min_value=hist_min,
-            max_value=hist_max,
             one_bin_per_int=True,
             split_x_axis_at= crosslinker_length if accepted_deviation_upper_bound is None else crosslinker_length+accepted_deviation_upper_bound
         )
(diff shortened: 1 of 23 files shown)


backend/protzilla/data_analysis/crosslinking_validation.py
```diff
@@ -931,3 +974,200 @@ def multimer_diagrams(
...
+    if lin_end <= lin_start:
+        lin_end = lin_start + 1
+    lin_bins = np.arange(lin_start, lin_end + 1, 1)
+
+    # 2. Log Bins
+    # We must ensure max_distance > split_x_axis_at to avoid math domain errors
+    safe_max = max(max_distance, split_x_axis_at * 1.01)
+    log_start = np.log10(split_x_axis_at)
+    log_end = np.log10(safe_max)
+    log_bins_transformed = np.arange(log_start, log_end + 0.1, 0.1)
```
tE3m, line 1055: This fails if `split_x_axis_at` is 0, since the logarithm of 0 evaluates to `-inf`

backend/protzilla/data_analysis/crosslinking_validation.py
```diff
@@ -716,46 +772,50 @@ def diagrams_of_crosslinking_validation_data(
...
-            visual_transformation="linear",
-            relevant_column_a="alphafold_distance",
-            relevant_column_b="alphafold_distance",
-            one_bin_per_int=True,
+            xaxis_label="Distance in Å",
+            yaxis_label="Count",
+            split_x_axis_at=(
+                crosslinker_length
+                if accepted_deviation_upper_bound is None
+                else crosslinker_length + accepted_deviation_upper_bound
```
tE3m, line 787: related to the other comment: `accepted_deviation_upper_bound` is never `None` because the frontend enforces a numeric value (0 by default)

backend/protzilla/data_analysis/crosslinking_validation.py
```diff
@@ -716,46 +772,50 @@ def diagrams_of_crosslinking_validation_data(
...
+            title_invalid=f"Predictions not matching CLs (intra: {invalid_intra}, inter: {invalid_inter})",
             heading=f"Predicted distances for {structures_to_validate_str} with crosslinker {crosslinker}",
-            x_title="Distance (Å)",
-            y_title="Count",
-            overlay=True,
-            visual_transformation="linear",
-            relevant_column_a="alphafold_distance",
-            relevant_column_b="alphafold_distance",
-            one_bin_per_int=True,
+            xaxis_label="Distance in Å",
```
tE3m, line 782: is this actually better than "Distance (Å)"?

tE3m (changes requested):
Take a look at the x_axis cutoff, apart from this error the changes look good

backend/protzilla/data_analysis/crosslinking_validation.py
NeleRiediger: It actually bothered me that the split diagram now had "Distance in Å (linear/ log)" and the diagram above still had the "Distance (Å)" spelling. So for consistency reasons I actually like the new version better. 
(I was contemplating whether I want to mention it, because it is really petty. But since it was already mentioned here I had something to say to it :))

backend/protzilla/data_analysis/crosslinking_validation.py
```diff
@@ -931,3 +974,200 @@ def multimer_diagrams(
...
+        (~validated_df["valid_crosslink"]) & (validated_df["link_type"] == "inter")
+    ).sum()
+
+    return create_bar_plot(
+        values_of_sectors=[
+            valid_crosslinks,
+            invalid_crosslinks,
+        ],
+        names_of_sectors=[
+            f"Cross-Links matching predicted data (intra: {valid_intra_total}, inter: {valid_inter_total})",
```
NeleRiediger: Since we just agreed to change Cross-Links to Crosslinks, it would be nice to directly change that here

backend/protzilla/data_analysis/crosslinking_validation.py
```diff
@@ -931,3 +974,200 @@ def multimer_diagrams(
...
+            f"Cross-Links not matching predicted data (intra: {invalid_intra_total}, inter: {invalid_inter_total})",
```
NeleRiediger: Same here

backend/protzilla/data_analysis/crosslinking_validation.py
```diff
@@ -931,3 +974,200 @@ def multimer_diagrams(
...
+        ],
+        heading=f"All Cross-Links used for validation of {structures_to_validate_str}",
```
NeleRiediger: and here

backend/protzilla/data_analysis/crosslinking_validation.py
```diff
@@ -931,3 +974,200 @@ def multimer_diagrams(
...
+        y_title="Number of Cross-Links",
```
NeleRiediger: and here

NeleRiediger (commented):
Testing worked as expected (apart from the error @tE3m already mentioned, when the input fields are left at zero). 
Other than that maybe have a look at the consistency of the x-axis naming, but it is also absolutely not necessary to change anything, it's just about perfectionism. 
Code looks good to me too. I just wrote comments regarding the spelling of "crosslinking", that should ideally be resolved before merging.

backend/protzilla/data_analysis/crosslinking_validation.py
jorisfu: Added a fix for safety, but this should only happen if we have a CL with len = 0, which would be kinda odd IMO.

backend/protzilla/data_analysis/crosslinking_validation.py
jorisfu: Agree generally but this is better for safety and does not rely on how the front-end works IMHO

backend/protzilla/data_analysis/crosslinking_validation.py
AnnaPolensky: I originally changed it because it bothered me to have two brackets in a row ("(log) (A)"). 
I changed it for the other diagram (histogram_two_standard_deviations) as well now.

backend/protzilla/data_analysis/crosslinking_validation.py
tE3m: it's not exactly independent of how the frontend works since this will still fail if `crosslinker_length` is 0 for some reason, but we can leave as is

tE3m (approved):

jorisfu (approved):

PR #396, opened 2026-05-04
376 default values for cl validation

### LINKED ISSUE #376: Form for CL validation fetches form from disk to have dynamic fields work on runner
## Issue description
The current implementation has dynamic fields that only exsist if the CL data is already imported. This however will most likely not work with the runner, as one has to fill out the parameters before running it. The fields will not exist at that point. We therefore need the fields to exist even before we import the CL data.

## Acceptance Criteria
- [ ]

Elena-kal:
## Description
fixes #376 
Before we could only fill in crosslink length, upper and lower deviation after we imported the crosslinking data. Now we can add defaults for them in settings. These values are then automatically used for the crosslink with the same name.

## Changes
I decided to keep the default setting general to be able to work with it for other usages. Therefore one can find the default adding and reading in the same file as the disk_operator. (disk_operator.py) I then use it in views_helpers to create defaults for crosslinks. I also added the tab for the defaults in settings and changed the form for crosslinking validation slightly to take the set defaults as (...well) default.

## Testing
Set a default for DSSO in settings and then create a workflow and check whether the default values you set are then in the corresponding fields in crosslinking validation. Check both multimer and monomer. You can also delete the defaults and see whether the behavior then is as you would expect. You can of course also update the values and see whether everything behaves as expected.

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**

### DIFF
--- a/backend/protzilla/form.py
+++ b/backend/protzilla/form.py
@@ -242,10 +242,6 @@ class Form:
         if new_field.name in self._value_buffer:
             new_field.value = self._value_buffer.pop(new_field.name)
 
-    def add_field(self, new_field: InputField):
-        self.input_fields.append(new_field)
-        self._field_map[new_field.name] = new_field
-
     def __getitem__(self, fieldname: str) -> InputField:
         "to do form[fieldname] to get the field object"
 
--- a/backend/protzilla/methods/data_analysis.py
+++ b/backend/protzilla/methods/data_analysis.py
@@ -2377,11 +2377,18 @@ class CrosslinkingValidationWithAngstromStep(DataAnalysisStep):
         for crosslinker in crosslinkers:
             field_name = f"{crosslinker}_length"
             if field_name not in form:
-                cl_defaults = run.disk_operator.defaults.read_default(crosslinker)
-                if cl_defaults:
-                    length_default = cl_defaults["cl_length"]
-                    upper_deviation_default = cl_defaults["cl_upper_deviation"]
-                    lower_deviation_default = cl_defaults["cl_lower_deviation"]
+                cl_defaults = (
+                    run.disk_operator.defaults.read_default("crosslinker_lengths") or {}
+                )
+                specific_cl_defaults = cl_defaults.get(crosslinker, {})
+                if specific_cl_defaults:
+                    length_default = specific_cl_defaults.get("cl_length")
+                    upper_deviation_default = specific_cl_defaults.get(
+                        "cl_upper_deviation"
+                    )
+                    lower_deviation_default = specific_cl_defaults.get(
+                        "cl_lower_deviation"
+                    )
                 else:
                     length_default = 0
                     upper_deviation_default = 0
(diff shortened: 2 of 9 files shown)


backend/protzilla/methods/data_analysis.py
```diff
@@ -95,13 +95,6 @@
     monomer_validation,
     multimer_validation,
 )
-from backend.protzilla.run import Run
-from backend.protzilla.methods.importing import (
-    ImportMonomerStructurePredictionFromDisk,
-    AlphaFoldPredictionLoad,
-    ImportMultimerStructurePredictionFromDisk,
-    UploadMultimerPredictions,
-)
```
jorisfu, line 104: what happened here?

backend/protzilla/form.py
```diff
@@ -242,10 +242,6 @@ def add_field(self, new_field: InputField) -> None:
         if new_field.name in self._value_buffer:
             new_field.value = self._value_buffer.pop(new_field.name)
 
-    def add_field(self, new_field: InputField):
-        self.input_fields.append(new_field)
-        self._field_map[new_field.name] = new_field
```
jorisfu, line 247: why is this gone?

jorisfu (changes requested):
Code looks good apart from some questions. If we want this new defaults file to be general-purpose, the CL lengths should be under some sub-key, not just at the root level. E.g. we could to crosslinker_lengths->dsso->(etc)

backend/protzilla/form.py
Elena-kal: It is repetitive, directly above it is another add_field

backend/protzilla/methods/data_analysis.py
Elena-kal: unused imports

jorisfu (approved):

AnnaPolensky (changes requested):
The code looks good and I really like the functionality. However, three things:

- I looked at the failing backend pipeline. Two of your tests are failing (test_get_cl_defaults, test_update_cl_default_success)
- Since Nele just did a PR where she corrected all frontend facing spelling of "crosslinking", I think it would be nice if you could also change the spelling in the frontend parts you just added.
- Is it intended that you can only have a default for exactly one crosslinker? I think we should allow defaults for all the different crosslinker types you might use. Especially since more than one crosslinker type might be used for the same validation.

Elena-kal:
> The code looks good and I really like the functionality. However, three things:
> 
> * I looked at the failing backend pipeline. Two of your tests are failing (test_get_cl_defaults, test_update_cl_default_success)
> * Since Nele just did a PR where she corrected all frontend facing spelling of "crosslinking", I think it would be nice if you could also change the spelling in the frontend parts you just added.
> * Is it intended that you can only have a default for exactly one crosslinker? I think we should allow defaults for all the different crosslinker types you might use. Especially since more than one crosslinker type might be used for the same validation.

Hey Anna,
sorry for the bug. It sneaked in when implemented the changes Joris suggested, same with the tests. But I should have checked that, so sorry for the additional work. 
I believe it should be fixed now.

AnnaPolensky (approved):
Sorry for taking so long, I did not see that I had to review again. But it works now as expected 👍

AnnaPolensky:
One more question, did you create an issue for changing the color of the default icon when clicking on it?

Elena-kal:
> One more question, did you create an issue for changing the color of the default icon when clicking on it?

No not yet, but will do now.

PR #410, opened 2026-05-07
fix bug which prevents showing predictions in tables in frontend

### LINKED ISSUE #403: fix multimer predictions showing up in settings
Regarding the uploading step for AlphaFold multimer predictions:
If you don't enter the AlphaFold model while uploading, the uploaded multimer predictions won't show up in settings.
This bug also deletes already uploaded models from the settings.
Test also for the other non-required fields, that the model will show up in the settings if these fields are empty.

Elena-kal:
## Description
fixes #403 
Fixes the bug described in the issue. 

## Changes
The "required" fields are now consistent in forms and settings. Also fixes the bug by turning NaNs into empty strings.

## Testing
Upload a monomer or multimer prediction in forms or settings and leave a non-required field empty. Then look into the settings and the table where uploaded structures are displayed. You should be able to see your uploaded structures in the table there.

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**

### DIFF
--- a/backend/main/views_settings.py
+++ b/backend/main/views_settings.py
@@ -355,6 +355,7 @@ def get_monomer_structure(request):
     ]
 
     df = get_metadata_df(csv_file_path=metadata_csv, expected_columns=expected_columns)
+    df = df.fillna("")
 
     df_infos = df.rename(
         columns={
@@ -470,6 +471,7 @@ def get_multimer_structure(request):
         "model_used",
     ]
     df = get_metadata_df(csv_file_path=metadata_csv, expected_columns=expected_columns)
+    df = df.fillna("")
 
     df_infos = df.rename(
         columns={
--- a/backend/main/views_settings.py
+++ b/backend/main/views_settings.py
@@ -248,6 +248,7 @@ def check_and_copy_files_to_directory(file_names: list, target_dir: str):
         source_file = settings.FILE_UPLOAD_TEMP_DIR / file_name
         success, message = copy_file_to_directory(source_file, target_dir)
         if not success:
+            shutil.rmtree(target_dir, ignore_errors=True)
             return False, message
     return True, "All files successfully uploaded"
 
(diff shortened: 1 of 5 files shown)


jorisfu (changes requested):
TL;DR: Make required fields actually required

Leaving non-required fields empty does seem to work. However, leaving required fields empty also works. This leads to the following bugs I spotted in settings:

- Leaving an empty entry ID does create a new entry that cannot be deleted because we cannot actually create a folder with no name, hence we cannot delete a folder with no name and the deletion fails
- You can leave out any of the files and still create/delete the prediction, but you then get an error when importing it due to missing files. I think it'd be reasonable to check whether or not the files exist when uploading (validation can be deferred to the importing step IMO)
- You can leave out the Protein IDs and you won't get any error when uploading the prediction. You also won't get any error when importing it, but the `uniport_ids` in the `structure_metadata_df` will then be `[*]` (silent error I suppose)

And in the upload step:
- Leaving protein IDs empty results in "" in the `structure_metadata_df` (silent error)
- Protein IDs is not marked as required in the step form
- No AlphaFold model does not resolve to NaN but "" when using the upload step as opposed to settings. I think this is better than NaN and should be the default for no entered model
- New folders for persistent imports are always created, both if the step fails (due to missing files e.g.) or not, leading to empty folders that do however show up in settings as valid predictions

Elena-kal:
Thank you for your thorough review.

I have now fixed/addressed the following:
- One can no longer leave empty the entry id nor the uniprot ids
- all uniprot ids fields are marked as required now
- the entry no longer is created although some files are missing. Instead we first upload all files and then go through with the rest, if an error occurs afterwards, the folder is deleted and the upload fails with an error message shown
- I was not entirely sure what you meant with the default to NaN, because I think writing an empty string into a csv turns it into a NaN anyway, but I believe I addressed it with `"" if model_used is None else model_used`. If this is not what you meant please clarify.
- If the import is not successful we always delete the folder so we won't end up with new unused empty folders.

Again, thanks for the review. I think this saved us from quite a few bugs down the line.

jorisfu (approved):
works as expected now, thx for the fixes

PR #411, opened 2026-05-08
fix bug that causes user to have to press the download button twice

### LINKED ISSUE #353: "Download as CSV" button needs to be clicked two times to start download
# What is the issue?
See title

# How to reproduce the issue?
Take any table. Press the button. Nothing happens. Press again. CSV now downloads

The CSV download should occur on the first button press

Elena-kal:
## Description
fixes #353 
Before the user needed to press the download as csv button twice. Now it works as intended and expected, where one only needs to press it once.

## Changes
Before we used `setData` to update the variable `data`. However, the data variable is not updated within the calling of the function but only after it ran through it. This is why when we then checked for `if (data.length === 0)`, data was always empty and therefore we always returned. After the call, data was not empty anymore causing it to work in the second run. Now we store it in a local variable, as there is no need to use a state like this.

## Testing
Go to any workflow and press the 'Download as CSV' button to download a table. It should download it immediately.

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**

### DIFF
--- a/frontend/src/components/core/shared/button/button.tsx
+++ b/frontend/src/components/core/shared/button/button.tsx
@@ -560,27 +560,27 @@ export const CSVButton: React.FC<CSVButtonProps> = ({
   ...params
 }) => {
   const [isLoading, setLoading] = useState(false);
-  const [data, setData] = useState([]);
 
   const downloadCSV = async () => {
     setLoading(true);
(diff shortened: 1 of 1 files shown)


tE3m (approved):
If it walks like a fix and it talks like a fix... then Elena did some solid bug-finding! 🦆

PR #412, opened 2026-05-11
add option to show outliers or exclude them from view for all boxplots

### LINKED ISSUE #350: Allow for boxplots without outliers highlighted
## Issue description
Currently, all boxplots created via `create_box_plots` show all outlying data points as individual dots. This significantly slows down the front-end and is _most likely_ not always helpful for readability either, especially for large datasets.

To improve performance and readability, we should add another plot option (e.g. "Highlight outliers") with the corresponding options that plotly supports so that we can call `go.Box` with `boxpoints=False` per default.

## Acceptance Criteria
- [ ] Steps that output a boxplot have a plot option that allows to change the `boxpoints` attribute of `go.Box` to the supported values by plotly
- [ ] The default value of this is `False`

Elena-kal:
## Description
fixes #350 
Every step that has a boxplot as output has the option to not show outliers.

## Changes
Changes were made to the create_boxplot function but the effects ripple through all plotting functions that use it. I also added the new checkbox field in the corresponding forms.

## Testing
Create a workflow that includes normalisation or imputation and check out the boxplots with and without outliers.

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**


jorisfu (commented):
Works fine from a technical perspective. However, the plot outputs are kind of weird: When choosing to not show outliers, the whiskers extend to the maximum outlier point instead of where they ended previously. I'm not sure this is how boxplots are supposed to work, but I suppose the plotly devs did this for some reason.

Also, this doesn't slow down loading times because the effective plot size remains the same. Rendering less points does speed up the front-end and makes it more laggy though, so I think this is still an improvement.

I think we should wait for Chris' approval tomorrow on this

jorisfu (changes requested):
Works, but as discussed in today's meeting we should specify "standard statistical boundaries" to the actual boundaries used by plotly. Also I'd be great if we could find a way to extract that info field to one constant so we can reuse that instead of repeating code

jorisfu (approved):

PR #420, opened 2026-05-13
fix: start frontend only after backend is ready

tE3m:
## Description

fixes the frontend not receiving any data after restarting the dev services (e.g. no workflows/runs). This has the side effect of the frontend not starting if the backend server runs into errors while starting up, which is better than being able to reach an empty frontend imo. Also disables frozen modules to get rid of the debugpy warning

## Changes

- add a simple endpoint at `/healthcheck/` that replies with a plain `HttpResponse` (`urls.py` and `views.py`)
- add a `healthcheck` to the compose service that tries to access this endpoint
  - add the condition that the frontend only starts once the backend is healthy (i.e. reachable)

## Testing

Start ProtZILLA as you usually would, open the home screen in the browser. Then, stop the services but keep the tab open. Start the services again and focus the browser tab; before, this would lead to the frontend eagerly trying to retrieve data from the backend before it was ready. Now, there should be no errors in the logs and the home screen should only show up if there is also data to populate the runs table and workflows row

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**

### DIFF
--- a/backend/Dockerfile.dev
+++ b/backend/Dockerfile.dev
@@ -16,4 +16,4 @@ RUN --mount=type=bind,source=install_scripts/database_download.py,target=install
     python install_scripts/database_download.py
 
 ENV DEBUGMODE=1
-ENTRYPOINT ["bash", "-c", "python -m debugpy --listen 0.0.0.0:5678 backend/manage.py runserver 0.0.0.0:8000 --nothreading"]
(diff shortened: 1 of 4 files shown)


Elena-kal (approved):
Works, and code looks good. Well done :))

PR #424, opened 2026-05-17
313 add server side sorting and filtering of tables

### LINKED ISSUE #313: Add Server-Side Sorting and Filtering of Tables
## Issue description
Tables cannot be sorted in the front-end, as they are not fully loaded. We need support to sort the tables by specific columns using the back-end. Also, table columns must be filterable from the frontend and return correct data from the backend

## Acceptance Criteria
- [ ] Tables can be sorted by column values in the front-end
- [ ] Sorted tables are correctly retrieved from the back-end, we don't just sort the currently visible rows
- [ ] Same thing for filtering

AnnaPolensky:
## Description
fixes #313 
Added server-side filtering and sorting.
grafik
You can filter only one column at a time because filtering on two columns is available only in the MUI X Data Grid Pro Edition, not in the community edition.

## Changes

- added sorting and filtering parameters to get_current_step_table_data in views.py
- renamed _step_output_as_serialised_table to _dataframe_as_datagrid_rows and moved it from views.py to views_helper.py
- frontend changes for the new parameters, mainly in data-table.tsx

## Testing

- do an import (e.g. MaxQuantImport)
- check that you can sort based on the different columns
- check different filters (maybe for strings and for numbers)
- check filtering and sorting for tables that are based on lists (like contaminants in the MaxQuantImport) as well
- filter and sort a list, download it and check that the download matches the filtered down and sorted version displayed

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**

### DIFF
--- a/backend/main/views.py
+++ b/backend/main/views.py
@@ -683,62 +683,18 @@ def get_step_plots(request):
         )
 
 
-# # TODO: Move somewhere else
-# def _step_output_as_serialised_table(
-#     label: str, _data: pd.DataFrame | Any, index_delims: tuple[int, int] = (None, None)
-# ) -> list[dict]:
-#     """
-#     Returns the output data of a step as a list of dicts in "records" orientaion, like this:
-#     [{'col1': 1, 'col2': 0.5}, {'col1': 2, 'col2': 0.75}]
-#     Also delimits the return according to index_delims.
-#     If the output could not be serialised, None is returned
-#
-#     :param label: The label of the step output to serialise
-#     :param _data: The data associated with the output
-#     :param index_delims: tuple used as slice begin and end indices to delimit the output
-#     """
-#     start_index = index_delims[0]
-#     end_index = index_delims[1]
-#
-#     # Note: using [None:None] as a slice returns the entire collection
-#     if isinstance(_data, pd.DataFrame):
-#         data = _data.iloc[start_index:end_index].copy()
-#
-#         # Safer than just adding the new column. We assume __id_col is not
-#         # a column name anyone would use
-#         if "id" in data.columns:
-#             data.rename(columns={"id": "__id_col"}, inplace=True)
-#
-#         data["id"] = data.index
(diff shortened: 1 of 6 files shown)


frontend/src/components/core/shared/button/button.tsx
NeleRiediger: I don't know if there is a useful replacement for this "any"? But since there is a Linter warning I feel like we should try to replace it or at least write a specific comment why "any" is the best option here.

NeleRiediger (approved):
Everything I tested worked fine and code looks good too.

frontend/src/components/core/shared/button/button.tsx
AnnaPolensky: Introduced a new interface for TableRows to remove the any.

backend/main/views.py
```diff
@@ -789,9 +751,44 @@ def get_current_step_table_data(request):
...
+            value = f.get("value")
+
+            if not field or value is None:
+                continue
+
+            col = step_output[field].astype(str)
+
+            step_output = step_output[
+                col.str.contains(str(value), case=False, na=False)
+            ]
```
jorisfu: This doesn't allow for numeric filtering, e.g. `>` or `<` that the front-end allows for numerical columns

jorisfu (changes requested):
Technically well implemented but we need consistency between the operators allowed for filtering in the front-end and the supported operations in the back-end, e.g. actually check the "operator" key in the filtering metadata and apply it or remove some filtering operators from the front-end that we don't need.

I don't think we need all of the string ops but numerical stuff (>, <, !=) should be supported

backend/main/views.py
AnnaPolensky: Thank you for pointing this out. Limited the string operator selection to contains and equals and the numeric operator selection to >,<,=. They are now also implemented in the backend.

jorisfu (approved):
works 👍

PR #431, opened 2026-05-20
311 fix run deletion

### LINKED ISSUE #311: BUG: Run deletion fails if Run can't be instantiated
# What is the issue?

If a run can't successfully be instantiated, e.g. due to the run.yaml containing invalid/unknown tags that can't be deserialized, the run won't be deleted. However, it will disappear from the frontend until reload.

# How to reproduce the issue?

Create a run.yaml that contains errors and try loading it. This bug only appears if there is a run with a corresponding name already in the server's known instances, so runs that are purely on disk aren't affected.
After attempting to load it, return to the main menu. There, click on the delete button of the run. It will disappear upon clicking but won't actually be deleted. Reload and see it still present.

Deleting shouldn't silently fail. It should especially not erroneously show success (disappearance of the run upon button click).
First, the deletion should be a synchronous action, the run shouldn't disappear unless it has actually been deleted or the attempt has failed. For this, the server should eventually respond with a success/failure HTTP code.
Secondly, if there are errors, display them to the user!
Thirdly, deletion of the run shoudn't fail at all if the run can't be successfully loaded. Oftentimes, erroneous runs are exactly the ones you'd like to delete.

# Files

I expect changes in `backend/main/views.py` where the `delete_run` method is defined and in `frontend/src/components/app/runs-table/runs-table.tsx` in the `handleDeleteRun` method

Elena-kal:
## Description
fixes #311 
The bug is well-documented in the original issue, so I will focus on the fixes below.

## Changes
As @tE3m already correctly identified, deletion was failing silently because the server tried to instantiate the run first, which breaks if the run is corrupted. 
Now, if instantiation fails, we catch the error and force-delete the run directly, bypassing the need to instantiate the run in order to call delete_run. 
I also changed the error message sent to the frontend slightly to better reflect what is actually going wrong. 
Lastly, I also changed how the frontend handles deletions. Instead of expecting the deletion to work and already update the run table accordingly, we wait for the run to be successfully deleted. Only after that we actually update the table. 
If the deletion fails, we now display an error for the user and don't fail silently. 

## Testing
Create a run while on the crosslinking branch and create some steps that are not on dev (e.g. the multimer importing step). Next, switch the branch to `311-fix-run-deletion`. Here, try to delete the run. Reload and it should still be deleted. Create a new run with the same name, it should be a new run and not the old one again. 
To be honest, I am not sure how to test the error messaging because everything works :P 
But of course feel free to stress test and break things.

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**

### DIFF
--- a/backend/main/views.py
+++ b/backend/main/views.py
@@ -197,21 +197,25 @@ def delete_run(request):
         run_name = data.get("run_name")
 
         try:
-            if run_name not in Run._instances:
-                delete_run_folder(run_name)
+            if run_name in Run._instances:
+                try:
+                    run = Run(run_name)
+                    run.delete_run()
+                except Exception:
+                    Run._instances.pop(run_name, None)
+                    delete_run_folder(run_name)
             else:
-                run = Run(run_name)
-                run.delete_run()
+                delete_run_folder(run_name)
 
             return JsonResponse({"success": True, "message": "Deleted run"})
         except Exception as e:
-            traceback.print_exc()  # not sure if it still needs to be here
             return JsonResponse(
                 {
                     "success": False,
-                    "message": format_trace(traceback.format_exception(e)),
+                    "message": str(e),
+                    "traceback": format_trace(traceback.format_exception(e)),
                 },
-                status=404,
+                status=500,
             )
     else:
         return JsonResponse(
--- a/backend/main/views.py
+++ b/backend/main/views.py
@@ -198,6 +198,14 @@ def delete_run(request):
 
         try:
             if run_name in Run._instances:
+                # there are the following three cases:
+                # 1. The run is in memory and healthy: We successfully retrieve it and
+                #    use its delete method to clean up state and release file locks.
+                # 2. The run is in memory but corrupted: Instantiation fails,
+                #    so we catch the exception, forcefully evict it from the cache,
+                #    and wipe the folder directly.
+                # 3. (Handled by the else block) The run is NOT in memory: We bypass
+                #    instantiation entirely to save resources and directly wipe the folder.
                 try:
                     run = Run(run_name)
                     run.delete_run()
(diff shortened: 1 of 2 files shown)


backend/main/views.py
```diff
@@ -197,21 +197,25 @@ def delete_run(request):
...
         try:
-            if run_name not in Run._instances:
-                delete_run_folder(run_name)
+            if run_name in Run._instances:
+                try:
+                    run = Run(run_name)
+                    run.delete_run()
+                except Exception:
+                    Run._instances.pop(run_name, None)
+                    delete_run_folder(run_name)
```
AnnaPolensky, line 214: I had a look at the delete_run() method of runs. It looks like this:
```
def delete_run(self) -> None:
        delete_run_folder(self.run_name)
        self._instances.pop(
            self.run_name, None
        )  # remove instance from the class dictionary
```

So, the except is basically doing exactly the same as the try, apart from the different order? So, could you potentially just swap the order in delete_run or do something like:

```
def delete_run(self) -> None:
    try:
        delete_run_folder(self.run_name)
    finally:
        self._instances.pop(self.run_name, None)
```

AnnaPolensky (commented):
Works :)

backend/main/views.py
Elena-kal: The difference between what happens in try is the instantiation of the run which we then delete. This is our "normal" case. However, if the run is broken (because for example it contains a step from another branch that this branch does not know) the instatiation will fail and that is when we land in exception.
The exception essentially does the same thing only it does not need to create the run. I guess what we could do is to not use delete_run at all and simply do it by ourselves without the need to create it, but the Run._instances.pop(run_name, None) feels more like something we only want to do in an exception and not as the normal case. But that is simply my feeling idk what do you say?

Elena-kal: I guess what I was trying to say is that simply changing delete_run will not do the trick because this is not where the problem lies.

AnnaPolensky: Thank you for the explanation. It makes sense to me now, so I would keep your implementation as is.

tE3m: At first, I also found it confusing but I get why this is implemented this way. Maybe there should be comments explaining why we have these three cases, but not critical

tE3m (approved):
sorry for the delay. works well, even if the code section seems a bit confusing

PR #433, opened 2026-05-22
fix: pie chart coloring is now consistent and does not switch with ma…

### LINKED ISSUE #144: Pie Charts with consistent colours
## Issue description
Make sure the plots have consistent colouring (e.g. Filtered Proteins always red, non-filtered always blue). Right now the colors change depending on which group is larger. This issue applies to all pie charts, bar charts seem to be unaffected

Elena-kal:
…jority

## Description
fixes #144 
Before, the colors of the pie chart switched along with the majority. Blue was always the larger part. Now it stays consistent with the x values. 

## Changes
Mini change: Set sort to False and the values are not sorted by their y values / theirs shares anymore causing the coloring to stay the same as well.

## Testing
Simply create a standard workflow and look at the pie chart in the filter step. Play around with the values so that the share of protein kept goes from majority to minority and you will see that the color stays the same.

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**

### DIFF
--- a/backend/protzilla/data_preprocessing/plots.py
+++ b/backend/protzilla/data_preprocessing/plots.py
@@ -39,7 +39,7 @@ def create_pie_plot(
     )
 
     fig.update_layout(title={"text": f"<b>{heading}</b>"})
-    fig.update_traces(hovertemplate="%{label} <br>Amount: %{value}")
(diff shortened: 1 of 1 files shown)


AnnaPolensky (approved):
Works 👍

PR #456, opened 2026-06-10
fix off by one error in crosslink position calculation 

### LINKED ISSUE #453: BUG: off by one error in crosslink position calculation
# What is the issue?
When wie calculate the absolute position of a crosslink for its validation, we are off by one. If we use the position to calculate the exact amino acid from the whole protein sequence, this amino acid does not match the one that was originally desired in the data (for the XLinkX data format we actually see the peptide sequence with the exact amino acid marked). (Note that both the position within the peptide and the absolute position within the protein are 1-based)

# How to reproduce the issue?
Do a validation and get the crosslinker table. Then get the protein sequence from the fasta and get the amino acid at the crosslinker position (one letter). Then find the peptide sequence of the binding site of this crosslinker in the crosslinking data and see which letter is marked there. 

The amino acid at the position of the crosslinker we calculated should match the amino acid desired in the data.

NeleRiediger:
## Description
fixes #453 
Already explained in the Issue description. 

## Changes
- I fixed the off-by-one error. 
- Then I fixed the tests that used 0 as a value for 1-based positions (those tests worked before because of the bug). 
- I also updated a docstring stating that the CL position within peptides was 0-based, while this is clearly not the case as stated elsewhere. 
- Lastly I also added a test making sure the amino acid of the CL position within peptides on the peptide sequence is the same as the amino acid of the absolute CL position on the whole protein sequence. So this would hopefully be caught in the future. (I had to move the function in question one layer up for this) 

## Testing
I described in the issue description how to reproduce the bug and you can check whether everything is correct now the same way. 

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**

### DIFF
--- a/backend/main/views_helper.py
+++ b/backend/main/views_helper.py
@@ -290,8 +290,8 @@ def extract_relevant_crosslink_information(
     As well as a boolean for its validity and wether it is an intra or inter crosslink.
 
     :param crosslinking_df: DataFrame with columns
-        'crosslinker_position1',
-        'crosslinker_position2',
+        '1_based_crosslinker_position1',
+        '1_based_crosslinker_position2',
         'Chain_id1',
         'Chain_id2',
         'valid_crosslink',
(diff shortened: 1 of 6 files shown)


AnnaPolensky (approved):
Looks good and I was able to validate the positions. 

Small suggestion: Maybe we could include in the column names that the positions are 1-based.

3dot141592:
I reproduced the issue you described in #453, and everything appears to be correct. Code looks perfect :).

3dot141592 (approved):

NeleRiediger:
> Looks good and I was able to validate the positions.
> 
> Small suggestion: Maybe we could include in the column names that the positions are 1-based.

I adjusted the column names for "crosslinker_position1/2" to "1-based_crosslinker_position1/2" and "CL_position_within_peptide1/2" to "1_based_CL_position_within_peptide1/2". Maybe just sanity check if I messed anything up :)

AnnaPolensky (approved):
Looks good :)

PR #462, opened 2026-07-08
461 log normalization method

### LINKED ISSUE #461: Add option to use normalization method for log transformed data
## Issue description
Our current normalization by median method assumes that the data is raw and not log transformed (although we do it differently in the standard workflow which is fine for most use cases Chris said.). However, we currently have no way to normalize correctly after log transformation. A correct normalization after a log transformation subtracts instead of dividing. 

## Acceptance Criteria
- [ ] users have the option to use a normalization method that assumes log transformed data
- [ ] users can also choose normalization methods that assume "raw" data (not log transformed)

Elena-kal:
## Description
fixes #461 
Adds another version of the normalization by median method.

## Changes
Made changes to the method, added option to choose the method in forms and added and adapted tests.

## Testing
Create a standard workflow, run it with checking the box. Switch the order and run it without checking the box.

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**

### DIFF
--- a/backend/protzilla/data_preprocessing/normalisation.py
+++ b/backend/protzilla/data_preprocessing/normalisation.py
@@ -1,6 +1,7 @@
 import logging
 
 import pandas as pd
+import numpy as np
 from sklearn.preprocessing import StandardScaler
 
 from backend.protzilla.data_preprocessing.plots import (
@@ -51,6 +52,7 @@ def by_z_score(protein_df: pd.DataFrame) -> dict:
 
 def by_median(
     protein_df: pd.DataFrame,
+    log: bool,
     percentile=0.5,  # quartile, default is median
 ) -> dict:
     """
@@ -62,6 +64,8 @@ def by_median(
     :param protein_df: the dataframe that should be filtered in
         long format
     :type protein_df: pandas DataFrame
+    :param log: whether the data was log transformed before the normalisation or not
+    :type log: bool
     :param percentile: the chosen quartile of the sample intensities for
         normalisation
     :type percentile: float
@@ -84,17 +88,33 @@ def by_median(
     samples = protein_df["Sample"].unique().tolist()
     zeroed_samples = []
 
+    if log:
+        valid_mask = np.isfinite(protein_df[intensity_name])
+        valid_data = protein_df.loc[valid_mask, intensity_name]
+        global_median = valid_data.median()
+
     for sample in samples:
         df_sample = protein_df.loc[protein_df["Sample"] == sample,]
         quantile = df_sample[intensity_name].quantile(q=percentile)
 
-        if quantile != 0:
-            df_sample[f"Normalised {intensity_name}"] = df_sample[intensity_name].div(
-                quantile
-            )
+        if log:
+            if np.isfinite(quantile):
+                # without adding the global median our data would be zero centered and therefore one half would be
+                # negative which can lead to problems later down the workflow
+                df_sample[f"Normalised {intensity_name}"] = (
+                    df_sample[intensity_name] - quantile + global_median
+                )
+            else:
+                df_sample[f"Normalised {intensity_name}"] = 0
+                zeroed_samples.append(sample)
         else:
-            df_sample[f"Normalised {intensity_name}"] = 0
-            zeroed_samples.append(sample)
+            if quantile != 0:
+                df_sample[f"Normalised {intensity_name}"] = df_sample[
+                    intensity_name
+                ].div(quantile)
+            else:
+                df_sample[f"Normalised {intensity_name}"] = 0
+                zeroed_samples.append(sample)
         df_sample.drop(axis=1, labels=[intensity_name], inplace=True)
         scaled_df = pd.concat([scaled_df, df_sample], ignore_index=True)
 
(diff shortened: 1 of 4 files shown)


backend/protzilla/data_preprocessing/normalisation.py
```diff
@@ -84,17 +88,33 @@ def by_median(
...
-        if quantile != 0:
-            df_sample[f"Normalised {intensity_name}"] = df_sample[intensity_name].div(
-                quantile
-            )
+        if log:
+            if np.isfinite(quantile):
+                # without adding the global median our data would be zero centered and therefore one half would be
+                # negative which can lead to problems later down the workflow
+                df_sample[f"Normalised {intensity_name}"] = (
+                    df_sample[intensity_name] - quantile + global_median
```
tE3m: I'm assuming you did, but just to be certain: this is the behavior Chris wants for the normalization?
I tried changing the order (first doing the median normalization and then log-transformation) and the result:

Image

is the same as if we didn't correct for the global median during the normalization (i.e. `df_sample[intensity_name] - quantile` instead of `df_sample[intensity_name] - quantile + global_median`):

Image

Just for reference, this is the output with correction enabled:
Image

So if this is the behavior Chris wants ("the users should know this"), I feel there should at least be a warning (or another checkbox). These steps being non-commutative would be surprising to me as a user

backend/protzilla/data_preprocessing/normalisation.py
```diff
@@ -84,17 +88,33 @@ def by_median(
...
+                zeroed_samples.append(sample)
         else:
-            df_sample[f"Normalised {intensity_name}"] = 0
-            zeroed_samples.append(sample)
+            if quantile != 0:
+                df_sample[f"Normalised {intensity_name}"] = df_sample[
+                    intensity_name
+                ].div(quantile)
+            else:
+                df_sample[f"Normalised {intensity_name}"] = 0
```
tE3m: nitpick, code style: feels like the column name could be extracted to its own variable, the f-string definition is repeated for every of the 4 cases

backend/protzilla/methods/data_preprocessing.py
```diff
@@ -673,6 +683,14 @@ def create_form(self):
             ],
         )
 
+    def modify_form(self, run):
+        if self.form["log"].value:
+            self.form["visual_transformation"].value = (
+                VisualTransformations.LINEAR.value
+            )
+        elif not self.form["log"].value:
+            self.form["visual_transformation"].value = VisualTransformations.LOG10.value
```
tE3m: nitpick, code style: two comments on this - I always feel that it's more legible to first define the form fields as variables in the `modifiy_form` methods. Secondly, since the enums are `StrEnum` instances, there's no need to access `.value` - `VisualTransformations.LOG10 == VisualTransformations.LOG10.value` is `True`.
Both non-critical and up to you, added a proposal for convenience
```suggestion
        visual_transformation_field: DropdownField = self.form["visual_transformation"]
        log_field: CheckboxField = self.form["log"]
        if log_field.value:
             visual_transformation_field.value = (
                VisualTransformations.LINEAR
            )
        elif not log_field.value:
            visual_transformation_field.value = VisualTransformations.LOG10
```

tE3m (commented):
One larger question regarding the execution of the correction, apart from that just some nitpicks that are up to you. Solid work overall, I would've forgotten about the runner

backend/protzilla/data_preprocessing/normalisation.py
Elena-kal: Fair, I will talk to Chris about this. To be honest, I needed it to centered around the previous median for my batch effect correction algorithms because they could not handle the negative values which to be fair do not make a lot of sense when dealing with intensities. But you are right, this should be at least documented because I guess this is not the most natural behavior.

backend/protzilla/methods/data_preprocessing.py
```diff
@@ -673,6 +683,14 @@ def create_form(self):
             ],
         )
 
+    def modify_form(self, run):
+        visual_transformation_field: DropdownField = self.form["visual_transformation"]
+        log_field: CheckboxField = self.form["log"]
+        if log_field.value:
+            visual_transformation_field.value = VisualTransformations.LINEAR
+        elif not log_field.value:
+            visual_transformation_field.value = VisualTransformations.LOG10
```
3dot141592: This sets `visual_transformation` on every form update, so the user cannot manually change the dropdown. If this is intentional, the field should probably be hidden/disabled, otherwise this seems like a bug.

3dot141592 (changes requested):
Just as Tarek noted, I find it very counterintuitive that the log and normalization steps dont commute... Perhaps we could add a "shift variables around the median" checkbox, or make it a separate step.

Apart from that and the minor frontend bug, it looks good.

backend/protzilla/data_preprocessing/normalisation.py
Elena-kal: Took it out and will deal with the shifting/scaling in another issue :))

3dot141592 (approved):

Issue #465, opened 2026-07-15
Add a scale method step


Elena-kal:
## Issue description
We want a step where the user can set a min and a max value and the all intensity values of the proteins map between those two values

## Acceptance Criteria
- [ ] users can set min and max values
- [ ] all values stay between those two values
- [ ] the relative distance between the values stays the same

PR #467, opened 2026-07-16
465 add scaling step

Elena-kal:
## Description
fixes #465 
This issue introduces a new step in which the intensity values can be scaled. 

## Changes
Added the new step to transformations in data_preprocessing and added new tests.

## Testing
Create a standard workflow. Add the scaling step (I would assume one would usually want the scaling step after the normalization and log transformation). And check out how the data scales. The min value of the protein df given should map to the minimum value set in the parameters, same with the maximum values.

## PR checklist
**Development**
 
**Mergeability**
 
**Code review**

### DIFF
--- a/backend/protzilla/data_preprocessing/transformation.py
+++ b/backend/protzilla/data_preprocessing/transformation.py
@@ -85,7 +85,7 @@ def by_log(
     return dict(protein_df=transformed_df, peptide_df=transformed_peptide_df)
 
 
-def by_log_plot(
+def transformation_plot(
     protein_df, output_protein_df, graph_type, group_by, show_outliers=True
 ):
     if graph_type == "Boxplot":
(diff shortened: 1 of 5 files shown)


backend/protzilla/data_preprocessing/transformation.py
```diff
@@ -110,3 +110,62 @@ def by_log_plot(
...
+    max_old = intensity_column.max()
+
+    if min_old == max_old:
+        raise ValueError(
+            "Global max equals global min; cannot scale a constant dataset."
+        )
+
+    transformed_intensity_column = (
+        (intensity_column - min_old) / (max_old - min_old)
+    ) * (max_value - min_value) + min_value
```
jorisfu, line 138: Ik this is a classic formula but I did once add a `lerp` helper in `utilties.py` that you _could_ use here

jorisfu (changes requested):
Tested working, but I'd personally add an error when the input max is smaller than the input min. It should work anyway (with the data flipped), but I think most of the time this is not intentional, so I suppose we need at least a warning or an error.

Also we do already have a `lerp` helper function that could make the code cleaner by using `.apply()` on the df with that function. Not insisting on this though

NeleRiediger (approved):
Seems to work fine and code looks good as well

Issue #470, opened 2026-08-04
BUG: Checkbox value synchronizing across steps


AnnaPolensky:
# What is the issue?
Setting the checkbox can also set the checkbox of another step. This might create confusion about what exactly was calculate/the user might calculate a step differently than intended if one does not notice the change.

# How to reproduce the issue?
Add any step that has a checkbox two times. Set/Unset the checkbox, switch to the other step and see that the truth value of this checkbox has also changed.

# What is the expected behavior?
Checking a checkbox should affect only the current step, not any other steps.

PR #471, opened 2026-08-04
fix: stop synchronizing of checkbox field values between steps

AnnaPolensky:
## Description
fixes #470 
The checkboxes did synchronize since they all had the id "checkbox" hardcoded. To stop this, every checkbox gets now an id based on the step id.

## Testing
Add the MaxQuantImporting Step two times to a workflow. Check and uncheck the boxes and switch between the steps to see that the checkboxes of the different steps do not influence each other.


Elena-kal (approved):
Code looks fine, and it works well.

